Issue #2, opened 2020-10-08
Database ORM / Collection Models


JacksonMaxfield:
## Use Case
*Please provide a use case to help us understand your request in context*

Managing database schema + inserting new items into the database will be simpler with an ORM style API

## Solution
*Please describe your ideal solution*

I.E.:

```python
full_council = Event("Full Council", datetime(2020, 10, 7), "...")
database.insert(full_council)
```

PR #4, opened 2020-10-10
feature/file-person-body-seat-role-database-models

JacksonMaxfield:

WIP #2 

This adds database models for the `File`, `Person`, `Body`, `Seat`, and `Role` collections along with some basic validators for fields.

I have slightly changed the schema from our original document. Reasoning below.

---

### Major Changes:

Instead of including a `MapField` where we reference the id and duplicate values from the referenced foreign collection, I am just using a `ReferenceField`. As discussed in Slack, since we are dropping support for multiple database types we can optimize our queries a bit better and instead of doing things like `select_rows_as_list` stuff from `cdptools~=2` we can use the native Firestore API.

Our largest hit will be to large queries where we need to do joins on the front-end. An example of this would be when populating the voting records table. But with pagination and such I feel we can reduce the amount of requests we would need to make on a single page of the voting records table render.

---

### Field Changes:

`created` and `updated` fields have been removed from all collection models. To access `created` and `updated` fields as Python native `datetime` objects, after retrieving or saving a document you can use the model's `get_firestore_create_time()` function and convert to `datetime` with `ToDatetime()`:

Example:
```python
from cdp_backend.database.models import Body
from datetime import datetime

b = Body()
b.name = "Full Council"
b.is_active = True
b.start_datetime = datetime.utcnow()
b.save()

b_created = b.get_firestore_create_time().ToDatetime()
```

#### Collections

Change log of all field changes. All of these are "up for debate" in a sense. Really just tell me if I am wrong.<br>
[Original Planned Schema](https://github.com/CouncilDataProject/cdptools/blob/release-3/docs/document_store_schema.md)

##### File

No changes.

##### Person

* **Changed:** `picture_file` as a `MapField` changed to `picture` as a `ReferenceField`
* **Removed:** `is_council_president`. Can be queried from the `Role` collection.
* **Removed:** `most_recent_seat`. Cyclic dependency. Can be queried from the `Seat` collection.
* **Removed:** `most_recent_chair_body`. Cyclic dependency. Can be queried from the `Role` collection.
* **Removed:** `terms_serving_in_current_seat_role`. Can be queried from the `Role` collection.
* **Removed:** `terms_serving_in_current_committee_chair_role`. Can be queried from the `Role` collection.

##### Body

* **Removed:** `chair_person_id`. Some bodies don't have chair people and can be queried from the `Role` collection.

##### Seat

No changes.

##### Role

* **Changed:** `person` as `MapField` changed to `person` as `ReferenceField`
* **Changed:** `body` as `MapField` changed to `body` as `ReferenceField`
* **Changed:** `seat_id` as `TextField` changed to `seat` as `ReferenceField`

- [ ] Provide relevant tests for your feature or bug fix.

Tests are provided for the utility and validation functions. I will add tests for the database models when all comments on the models themselves are resolved.

I _may_ make a pull request to [FireO](https://github.com/octabytes/FireO) itself to add a `description` parameter to the base `Field` object that sets the fields `__doc__` property. This would mean that for each field would could add a docstring and thus more auto-doc generation.


tohuynh:
>* **Removed:** `is_council_president`. Can be queried from the `Role` collection.
>* **Removed:** `most_recent_seat`. Cyclic dependency. Can be queried from the `Seat` collection.
>* **Removed:** `most_recent_chair_body`. Cyclic dependency. Can be queried from the `Role` collection.
>* **Removed:** `terms_serving_in_current_seat_role`. Can be queried from the `Role` collection.
>* **Removed:** `terms_serving_in_current_committee_chair_role`. Can be queried from the `Role` collection.

We would query for this information if we query the `Role` collection using the `Person` reference and then adding some code on the front-end to get the information? This is fine if we only need this information on the Person details page. If we also need them for each person in the list of person cards, this would be slow?

>* **Removed:** `most_recent_seat`. Cyclic dependency. Can be queried from the `Seat` collection.

There is no cyclic dependency? Because there is no mention of a person in the `Seat` collection?

In general, how would filtering work if we want to filter by a reference field's details.? For example, filter `event` by the body's name?

cdp_backend/utils/resources/content-types-audio.csv
```diff
@@ -0,0 +1,158 @@
+Name,Template,Reference
```
tohuynh, line 1: What are these csv files being used for?

JacksonMaxfield: Used for looking up ISO media types: https://github.com/CouncilDataProject/cdp-backend/blob/feature/body-and-person-orm/cdp_backend/utils/file_utils.py#L12

archival stuff :shrug:

JacksonMaxfield:
@tohuynh Great comments! Thanks, they made me think hahaha.

> > * **Removed:** `is_council_president`. Can be queried from the `Role` collection.
> > * **Removed:** `most_recent_seat`. Cyclic dependency. Can be queried from the `Seat` collection.
> > * **Removed:** `most_recent_chair_body`. Cyclic dependency. Can be queried from the `Role` collection.
> > * **Removed:** `terms_serving_in_current_seat_role`. Can be queried from the `Role` collection.
> > * **Removed:** `terms_serving_in_current_committee_chair_role`. Can be queried from the `Role` collection.
> 
> We would query for this information if we query the `Role` collection using the `Person` reference and then adding some code on the front-end to get the information? This is fine if we only need this information on the Person details page. If we also need them for each person in the list of person cards, this would be slow?

Hmmmmm mentally thinking about the people page queries:

1. Get all person entities with `is_active == True` (or w.e. subset we want, end result is you get some amount of people `n`)
2. For each person, get all corresponding roles, `Roles` filter: `person isin ["person_ref_1", "person_ref_2", "etc"]`
3. Map roles to each person on the front-end
4. For each person, get seat info for the "Council Member" `Seat`: `documentId isin ["seat_ref_1", "seat_ref_2", "etc"]` ([See here](https://stackoverflow.com/a/58780369)). If that doesn't work then single doc lookups for `n` people
5. For each seat, get seat map from file store (worst case `n` file requests)
5. Compute tenure on front-end because you have all roles for selected members, i.e. len of n times each person has been a "Council Member"

Total requests / response with data and file store:
Best case: `(3  + (n * image file transfer))`
Worst case: `(2 + (n * single doc lookups) + (n * image file transfer))`

I really doubt that can be _that_ expensive?

**Side note:** Instead of storing a whole image file for the map, it may be nice to store the geo-json boundaries of the districts so that we can always control the map styling? But that adds more requests to some mapping API... :thinking: 

**Further side note:** if we go this route it will be crucial to have good reducers and tests on the front end API to handle all this data state mapping and storage and such.

> > * **Removed:** `most_recent_seat`. Cyclic dependency. Can be queried from the `Seat` collection.
> 
> There is no cyclic dependency? Because there is no mention of a person in the `Seat` collection?

Correct. No cyclic dependency but I think I meant to say: "Can be queried from the `Role` collection." But further, now that I thought about it, once you have all roles, you can actually just compute this on the front end.

> In general, how would filtering work if we want to filter by a reference field's details.? For example, filter `event` by the body's **name?**

Ahhhhh now _that_ is definitely why we had some fields exposed for sure. But, I think we can optimize queries a bit like above. The example of the `Event` from body can be `Event` from `body_ref`. We just need to store a bit more state / a lookup table of `body_name` to `ref` after we request the data.

tohuynh:
>I really doubt that can be that expensive?

It doesn't look expensive. But the more we do this (later we might add more filtering capabilities) the more we fall into the trap of filtering on the front-end. For example, as of right now, could only filter person by `is_active` on the back-end. Any filtering about roles, seats, datetime (users may want council members during these years) would require to gather the data on the front-end first?

>Further side note: if we go this route it will be crucial to have good reducers and tests on the front end API to handle all this data state mapping and storage and such.

Agree. We can also probably save data (like the `Seat collection`) in the app state to avoid re-querying the db for the seat details

>Ahhhhh now that is definitely why we had some fields exposed for sure. But, I think we can optimize queries a bit like above. The example of the Event from body can be Event from body_ref. We just need to store a bit more state / a lookup table of body_name to ref after we request the data.

I was thinking maybe when we filter events by body's name, we would have already have all the body refs? In that case we can do `isin` again, but it is limited to 10 refs in the array.

tohuynh (approved):

isaacna (commented):
The querying discussion you and To had make sense. Overall I think `ReferenceField` is a big improvement over storing the other document id's

PR #5, opened 2020-10-13
feature/add-model-uniqueness-validation

JacksonMaxfield:

Similar to existing `cdptools` utilities we need a function to ensure that the model we are about to upload is unique when compared to the other pieces of data. This adds that.

Example usage:
```python
from cdp_backend.database import models
from cdp_backend.database.validators import model_is_unique

b = models.Body.Example()  # or w.e. model instance you have
if model_is_unique(b):
    b.save()
```

I also updated the other validator function names to match this type of "part-of-speech" layout where if the function is to be used in a conditional, it should be "answering" that conditional.

Thanks for contributing!

### DIFF
--- a/cdp_backend/database/exceptions.py
+++ b/cdp_backend/database/exceptions.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from typing import Dict, List
+
+from fireo.models import Model
+
+###############################################################################
+
+
+class UniquenessError(Exception):
+    def __init__(self, model: Model, conflicting_results: List[Dict]):
+        self.model = model
+        self.conflicting_results = conflicting_results
+
+    @property
+    def pk_values(self):
+        return {pk: getattr(self.model, pk) for pk in self.model._PRIMARY_KEYS}
+
+    @property
+    def conflicting_ids(self):
+        return [r.id for r in self.conflicting_results]
+
+    def __str__(self):
+        return (
+            f"Uniqueness constraint failed for {self.model.collection_name}. "
+            f"Found {len(self.conflicting_results)} for values: {self.pk_values}. "
+            f"Conflicting IDs: {self.conflicting_ids}."
+        )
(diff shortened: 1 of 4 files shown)


cdp_backend/database/exceptions.py
```diff
@@ -0,0 +1,29 @@
...
+    @property
+    def conflicting_ids(self):
+        return [r.id for r in self.conflicting_results]
+
+    def __str__(self):
+        return (
+            f"Uniqueness constraint failed for {self.model.collection_name}. "
+            f"Found {len(self.conflicting_results)} for values: {self.pk_values}. "
+            f"Conflicting IDs: {self.conflicting_ids}."
+        )
```
tohuynh, line 29: For my understanding, you don't need to call `super().__init__()` because you provided your own `__str__` implementation?

cdp_backend/database/validators.py
```diff
@@ -4,12 +4,57 @@
...
+    ... b = models.Body.Example()
+    ... if model_is_unique(b):
+    ...     b.save()
+    """
+    # Initialize query
+    query = model.__class__.collection
+
+    # Loop and fill query for each primary key
+    for pk in model._PRIMARY_KEYS:
+        query.filter(pk, "==", getattr(model, pk))
```
tohuynh, line 45: To verify, this does what it is supposed to do in compounding the queries right?
You don't need to write `query = query.filter(pk, "==", getattr(model, pk))` ?

tohuynh (approved):

cdp_backend/database/validators.py
isaacna: Yeah good point, it doesn't seem like the function actually changes the state of the object itself https://github.com/octabytes/FireO/blob/faa354ba11a937b15571c5b43559b5d3ba3207c6/src/fireo/managers/managers.py#L222-L224

isaacna (approved):

cdp_backend/database/exceptions.py
JacksonMaxfield: Hmmmm may as well call it anyway. Just to be safe?

cdp_backend/database/validators.py
JacksonMaxfield: Good catch! Fixed!

PR #6, opened 2020-10-18
feature/auto-build-db-diagram

JacksonMaxfield:

I wanted to write a script that auto-checks for cyclic dependencies in our database schema as I think I see some coming up in future PRs / model additions. I quickly realized that this could be handled for us by many existing systems that handle directed graphs / graph builds and visualizers and figured I could two-for-one, generate a db diagram and also check for cyclic dependencies in one go with one such tools.

I first tried with [`pyreverse`](https://www.logilab.org/blogentry/6883) from [`pylint`](https://www.pylint.org/) but found it wasn't flexible enough. I noticed however that it (much like everything else in the DAG world) utilizes [`dot` / `graphviz`](https://graphviz.org/) under-the-hood and figured I would just write my own custom handler.

So that's what I did.

We now have a db diagram that is built as part of `make docs` (and `make gen-docs`).

image

Let me know if you want styling changed and such, it's finicky but possible.

Thanks for contributing!

### DIFF
--- a/cdp_backend/bin/create_cdp_database_uml.py
+++ b/cdp_backend/bin/create_cdp_database_uml.py
@@ -60,7 +60,7 @@ def _construct_dot_file(args: Args):
             "fillcolor": "white",
         },
         edge_attr={
-            "arrowhead": "none",
+            "arrowhead": "vee",
         },
     )
 
@@ -108,7 +108,6 @@ def _construct_dot_file(args: Args):
     for model_name, cls in inspect.getmembers(models, inspect.isclass):
         if model_name not in ["Model", "datetime"]:
             # Attach fields for each model by using the Example
-            fields = []
             m = cls.Example()
 
             # Construct DAG points
--- a/cdp_backend/bin/create_cdp_database_uml.py
+++ b/cdp_backend/bin/create_cdp_database_uml.py
@@ -46,7 +46,7 @@ class Args(argparse.Namespace):
 ###############################################################################
 
 
-def _construct_dot_file(args: Args):
+def _construct_dot_file(output_file: str):
     dot = Digraph(
         comment="CDP Database Diagram",
         graph_attr={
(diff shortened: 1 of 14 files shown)


cdp_backend/bin/create_cdp_database_uml.py
```diff
@@ -0,0 +1,141 @@
...
+            "bgcolor": "transparent",
+            "splines": "compound",
+        },
+        node_attr={
+            "shape": "record",
+            "style": "filled",
+            "fillcolor": "white",
+        },
+        edge_attr={
+            "arrowhead": "none",
```
tohuynh, line 63: I think adding an arrowhead `vee` or `normal` would add some visual cues to help us comprehend the diagram easier. So we don't have check whether the line is attaching to rectangle as a whole or to a particular field.

cdp_backend/bin/create_cdp_database_uml.py
```diff
@@ -0,0 +1,141 @@
...
+            node_label = f"{model_name} | {fields_as_text}"
+
+            # Attach as a complete node
+            dot.node(model_name, node_label)
+
+    # Second pass: Create DAG
+    for model_name, cls in inspect.getmembers(models, inspect.isclass):
+        if model_name not in ["Model", "datetime"]:
+            # Attach fields for each model by using the Example
+            fields = []
```
tohuynh, line 111: This line is not needed?

tohuynh (approved):
This is really cool. Was it your intent to create the diagram and then manually inspect for cycles, or is there suppose to be code in create_cdp_database_uml.py to automatically detect cycles?

cdp_backend/bin/create_cdp_database_uml.py
JacksonMaxfield: Good catch!

cdp_backend/bin/create_cdp_database_uml.py
JacksonMaxfield: Yea I debated this for a bit. I can change back `normal`. Makes sense.

JacksonMaxfield (commented):
Good catch on the `Digraph` instead of `DAG`. Will add in a step to check for cyclic deps.

cdp_backend/bin/create_cdp_database_uml.py
JacksonMaxfield: Changed to `vee`

isaacna (approved):

Issue #12, opened 2020-11-05
Add builds for all platforms


JacksonMaxfield:
## Use Case
*Please provide a use case to help us understand your request in context*

While originally setup thinking that these pipelines would only run on servers, civic tech happens can happen and be used by all platforms and users.... so actually making sure we do that would be a good idea.

## Solution
*Please describe your ideal solution*

Add windows and mac to build CI and fix any bugs.

## Alternatives
*Please describe any alternatives you've considered, even if you've dismissed them*

Issue #10, opened 2020-10-30
Migrate minimal event data and minimal session data over from cdptools


JacksonMaxfield:
## Use Case
*Please provide a use case to help us understand your request in context*

Standardize and version the acceptable data format we accept for pipelines with the pipeline code itself rather than scrapper side.

## Solution
*Please describe your ideal solution*

Migrate the work done by @isaacna on [cdptools](https://github.com/CouncilDataProject/cdptools/pull/159) to this repo.

Potentially change the name from `MinimalEventData` and `MinimalSessionData` to just `EventData` and `SessionData` as I think with the `Optional` keys, they are the full event data spec.

We should also make objects for `VoteData` and such, which are a part of the `Optional[List[MinutesItems]]`. Basically the entire data structure of all minimal data and optional data should be documented as the accepted data structure we accept for pipelines.

## Notes

This unfortunately will require a bit of copying the database ORM definitions unless we can think of a way to programmatically construct all of these models. I.E. run through the DB ORM definitions and created NamedTuples of every single model with the appropriate `Optional` tags. I think it's possible.

isaacna:
Note to self, will also be adding an `Optional[List[EventMinutesItem]]`field to the `EventData`class

isaacna:
Now that I think about it, using the existing `fireo` models might be difficult because some models have a bottom up reference, for example `Session.event_ref`. 

But if we want to output a single EventData object, we'll need to store `sessions` as part of the `EventData` object as a top down reference. Unless there's a workaround I'm missing, this would be a circular dependency. 

We could do something like typing `Event.sessions` as `List[str]` for document id references, and update them after the session documents are created? But at that point all the `EventData` wouldn't really be in a single object.

With this would it make more sense to create separate ingestion objects apart from the `fireo` models? There would have to be some logic that handles uploading the nested object data to the database in the proper order to not mess up references

JacksonMaxfield:
Yea that's a good catch. I think it just makes sense to define a bunch of `NamedTuple` objects. Especially for documentation purposes having the ingestion structure really well documented would be nice.

Also note we can add / check / validate a bunch of the ingestion data based off of just primary keys where as the fireo models are "full objects" i.e. a `Person` model has name, email, etc, but in the ingestion, the dev _could_ add those but they _could_ also just add the person's name and we would take care of the rest because of primary key.

PR #20, opened 2020-11-26
Adding ingestion objects for Event gather 

isaacna:
 - Resolves [#10]
 - Adding objects for that we will use for data ingestion in the event gathering pipeline. The List[`EventData`] will be the output of a `get_events` task/function 
 - I stacked the classes in what I thought was the most intuitive order
 - Decided to use separate `NamedTuple` object instead of the `fireo` models based on the discussion in the issue
- [ ] Provide relevant tests for your feature or bug fix.

Thanks for contributing!

### DIFF
--- a/cdp_backend/pipeline/ingestion_models.py
+++ b/cdp_backend/pipeline/ingestion_models.py
@@ -55,7 +55,7 @@ class MinutesItem(NamedTuple):
 
 class EventMinutesItem(NamedTuple):
     minutes_item_ref: MinutesItem  # top down
-    index: int
+    event_minutes_item_index: int
     decision: str
     external_source_id: Optional[str]
     vote: Vote
--- a/cdp_backend/pipeline/ingestion_models.py
+++ b/cdp_backend/pipeline/ingestion_models.py
@@ -1,0 +1,66 @@
+from datetime import datetime
+from typing import NamedTuple, Optional, List
+
+
+class File(NamedTuple):
+    uri: str
+    name: str
+    description: Optional[str]
+    media_type: Optional[str]
+
+
+class Matter(NamedTuple):
+    name: str
+    matter_type: str
+    title: str
+    external_source_id: Optional[str]
+
+
+class Person(NamedTuple):
+    name: str
+    router_string: Optional[str]
+    email: Optional[str]
+    phone: Optional[int]
+    website: Optional[str]
+    picture_ref: Optional[File]
+    is_active: Optional[bool]
+    external_source_id: Optional[str]
+
+
+class Vote(NamedTuple):
+    matter_ref: Matter
+    person_ref: Person
+    decision: str
+    in_majority: bool
+    external_source_id: Optional[str]
+
+
+class SessionData(NamedTuple):
+    session_datetime: datetime
+    session_index: int
+    video_uri: str
+    caption_uri: Optional[str]
+    external_source_id: Optional[str]
+
+
+class MinutesItem(NamedTuple):
+    name: str
+    description: str
+    matter_ref: Optional[Matter]
+    external_source_id: Optional[str]
+
+
+class EventMinutesItem(NamedTuple):
+    minutes_item_ref: MinutesItem  # top down
+    index: int
+    decision: str
+    external_source_id: Optional[str]
+    vote: Vote
+
+
+class EventData(NamedTuple):
+    body: str
+    event_datetime: datetime
+    sessions: List[SessionData]
+    external_source_id: Optional[str]
+    event_minutes_items: Optional[List[EventMinutesItem]]  # top down
--- a/cdp_backend/pipeline/ingestion_models.py
+++ b/cdp_backend/pipeline/ingestion_models.py
@@ -1,3 +1,6 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
 from datetime import datetime
 from typing import NamedTuple, Optional, List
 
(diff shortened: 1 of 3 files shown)


cdp_backend/pipeline/ingestion_models.py
```diff
@@ -0,0 +1,66 @@
...
+class EventMinutesItem(NamedTuple):
+    minutes_item_ref: MinutesItem  # top down
+    index: int
+    decision: str
+    external_source_id: Optional[str]
+    vote: Vote
+
+
+class EventData(NamedTuple):
+    body: str
```
JacksonMaxfield, line 62: If we are expanding the rest of the models into `NamedTuple` objects, why not expand all of them, `Body` included.

cdp_backend/pipeline/ingestion_models.py
```diff
@@ -0,0 +1,66 @@
+from datetime import datetime
```
JacksonMaxfield, line 1: For consistency with the rest of the repo can you add:

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-
```

to the top of this file.

JacksonMaxfield (commented):
Couple of things:

1. Need to make an `__init__.py` in this `pipeline` module. When I tried to `make docs` on the branch it didn't make anything for this module because that file was missing.
2. The docs for `NamedTuple` object apparently _sucks_. I looked into `TypedDict` which makes it look better but removes the enforceability of parameters. Then I looked into `dataclass` which also makes docs look nice but it also makes all parameters be required, even with `Optional` typing. I don't know if we have other options besides making our own object or sticking with `NamedTuple` as it gets the functionality correct but the docs look bad.
3. Considering that these objects are almost entirely copied from the models and given my comment in the code itself about `body` -> `body_ref`, does it not make sense to build off of @tohuynh's work and write a script to generate these ingestion objects for us.

cdp_backend/pipeline/ingestion_models.py
isaacna: 1. Will do, didn't check that make docs worked
2. I don't mind an alternative if `NamedTuple` leads to bad docs, but something like [this](https://chrisdown.name/2015/09/20/removing-namedtuple-docstrings-from-sphinx.html) could be a possible solution if we wanna keep using it
3. I'm not familiar with To's work, is he already writing a similar script?

cdp_backend/pipeline/ingestion_models.py
isaacna: Will do

cdp_backend/pipeline/ingestion_models.py
JacksonMaxfield: 2. Ahhh that may be an interesting package to use for sure. May be worth checking out prior to more complex stuff like the following
3. Here is a link to the issue and comment explaining what I mean: https://github.com/CouncilDataProject/cookiecutter-cdp-deployment/issues/2#issuecomment-726279208

isaacna: Sounds good, I'll look into the script and see if I can make something similar for the ingestion objects

tohuynh: I think Jackson is saying to iterate through the model's fields to dynamically create the object. Is that right?

JacksonMaxfield: Yea that's what I am thinking. Basically just make the file every time using the models.

isaacna: If we generate from the database models, how do we want to handle the bottom-up vs top-down reference discussed in #10? 

Some fields in the database models have bottom-up reference, such as `Vote -> event_minutes_item_ref` but in our ingestion models it makes more sense for it to be the other way around such as  `EventMinutesItem -> vote_ref`. 

For example, in the database models DAG, `vote` points to `matter`, `event_minutes_item`, `event`. If we consider ingestion with `Event` being our single return object and create a DAG, we'd want to reverse the `vote -> event_minutes_item, event` edges, but still keep the `vote -> matter` edge right? 

Am I just overthinking this or would we not be able to just generate based solely on the database models? Because it doesn't seem like there's a consistent way to just flip the direction of the references without hardcoding in some exceptions.

JacksonMaxfield: Oh hm right. Yea I guess can you just try fixing the other commented items (i.e. making a `body` obj instead of just `body` as a `str`) and making the docs look nice somehow?

isaacna: Yep, I removed the ugly/unecessary "Alias" lines and I think the docstrings look decent now

isaacna:
I'll add tests for the models sometime later this week

docs/conf.py
```diff
@@ -183,3 +183,21 @@
...
+# -- Extra docstring configurations ------------------------------------
+def no_namedtuple_attrib_docstring(app, what, name,
+                                   obj, options, lines):
+    is_namedtuple_docstring = (
+        len(lines) == 1 and
+        lines[0].startswith('Alias for field number')
+    )
+    if is_namedtuple_docstring:
+        # We don't return, so we need to purge in-place
+        del lines[:]
```
JacksonMaxfield, line 196: FASINCATING. I didn't know you could do this.

JacksonMaxfield:
> I'll add tests for the models sometime later this week. For the purpose of unit testing, would we prefer having `Example` classmethods for the ingestion models like we do for the database models?

Hmmm I don't really think we need tests for these objects just example usage would be good.

isaacna:
I looked into `dataclass` a bit more, and I think it's overall better to use than `NamedTuple`:
1. It can be both mutable or immutable, which could be useful in the pipeline code if we need to do some object manipulation. It's also faster for reads and nesting
2. We can use `Optional` and also define default values for any params 
3. Even tho I ended up using the normal docstring markup, I think it looks better than the autogenerated stuff with `NamedTuple` 

Also I don't know why the code coverage is still failing after I included example usage, since the lines for the class attributes should be hit now

JacksonMaxfield:
Closing due to #26

PR #26, opened 2020-12-07
feature/ingestion-models

### LINKED ISSUE #10: Migrate minimal event data and minimal session data over from cdptools

JacksonMaxfield:

Resolves #10

Building off of @isaacna's work in #20, I first went through how we are currently accepting data in `cdptools` and then went through and figured out what we would need to add and where to fully utilize the new db schema.

I also edited the base db models and adjusted primary keys where I realized didn't make sense, set some parameters to required as we can either calc / generate them on the fly during the pipeline etc, and fixed documentation.

Finally, I added a bin script to generate documentation specifically for the `EventIngestionModel` class using the two examples I laid out in the `pipeline.ingestion_models` module (`EXAMPLE_MINIMAL_EVENT` and `EXAMPLE_FILLED_EVENT`).

To see documentation view of these examples:
```bash
git clone https://github.com/CouncilDataProject/cdp-backend.git
cd cdp-backend
git checkout feature/ingestion-models
pip install .[dev]
make docs
```

and navigate to "Event Data Ingestion Model".

Other random note: I made some comments about "generation of `router_string` on the person class", I looked up how to remove accent characters and seems very possible for us to create such a generator prior to document upload so should be good to go there. [StackOverflow example](https://stackoverflow.com/questions/4512590/latin-to-english-alphabet-hashing)

Pinging @nniiicc because I think you will like the auto-genned documentation for the ingestion model. Relates back to [Packaging Municipal Legislative Event Data for Reuse & Exchange](https://zenodo.org/record/3380592). Speaking of which, where is that poster? The abstract is in zenodo but not the poster and I remember I really liked it.

Thanks for contributing!


JacksonMaxfield:
Oh and @isaacna you were totally correct. There is really no way this could have been done programmatically top-down or something.

isaacna (approved):
Built the docs and the event ingestion model page looks great! I agree with making the `EventMinutesItem` optional so we can have a more minimal version of the data. Also agree with the other minor tweaks with the database and ingestion models. 

Also just out of curiosity, how did you get code coverage to count the ingestion model lines? Is it working now because the models are being used from the `test` subdirectory?

JacksonMaxfield:
> Also just out of curiosity, how did you get code coverage to count the ingestion model lines? Is it working now because the models are being used from the `test` subdirectory?

Yea I think its just because they are being used / initialized somewhere.

tohuynh:
I wasn't able to generate the docs. I got ` WARNING: Problem with "mdinclude" directive:
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe1 in position 1701: invalid continuation byte
` and then `WARNING: toctree contains reference to document 'event_data_ingestion_model' that doesn't have a title: no link will be generated`. Is this because of some encoding issue in event_data_ingestion_model.md?

isaacna:
> > Also just out of curiosity, how did you get code coverage to count the ingestion model lines? Is it working now because the models are being used from the `test` subdirectory?
> 
> Yea I think its just because they are being used / initialized somewhere.

Hm I guess it just doesn't working trying to create examples via classmethods

JacksonMaxfield:
@to

> I wasn't able to generate the docs. I got `WARNING: Problem with "mdinclude" directive: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe1 in position 1701: invalid continuation byte` and then `WARNING: toctree contains reference to document 'event_data_ingestion_model' that doesn't have a title: no link will be generated`. Is this because of some encoding issue in event_data_ingestion_model.md?

@tohuynh I just pushed a very small to change to explicitely in UTF-8, can you retry this branch and if it fails can I get the full stack trace?

cdp_backend/pipeline/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+
+"""Pipeline package for cdp_backend."""
+
+from .ingestion_models import EventIngestionModel  # noqa: F401
```
tohuynh, line 5: Could you also import the other ingestion models so they could be used in `cookiecutter-cdp-deployment` or it is not needed?

tohuynh (approved):
This looks good. And thanks for fixing the `make docs` error. It works for me now.

cdp_backend/pipeline/__init__.py
JacksonMaxfield: Hah. I think after reading this comment, I would rather just remove this shortener rather than make all of them shorter.

i.e. users would have to `from cdp_backend.pipeline.ingestion_models import EventIngestionModel` rather than current (`from cdp_backend.pipeline import EventIngestionModel`)

Keeping them all standard rather than making Event special makes more sense to me.

PR #29, opened 2020-12-17
fix/model-is-unique-return-type

isaacna:
- [ ] Link to any relevant issue in the PR description. Ex: _Resolves [gh-12], adds tiff file format support_
 - Ran into issues using the model_is_unique method while working on the event gather pipeline
 - Realized that the method was returning null, so you couldn't really use it like the doc described below:
```
if model_is_unique(b):
    b.save()
```
 - Made the method return a bool for easier use and added a test case
 - manually tested this with a fireo connection as well 

Thanks for contributing!


cdp_backend/database/validators.py
```diff
@@ -13,7 +13,7 @@
 # Model Validation
 
 
-def model_is_unique(model: Model) -> None:
+def model_is_unique(model: Model) -> bool:
```
JacksonMaxfield, line 16: Thinking a bit about this -- heres my hot take:

Maybe this function shouldn't raise at all? Just return a bool True / False?

cdp_backend/database/validators.py
```diff
@@ -49,6 +49,8 @@ def model_is_unique(model: Model) -> None:
     if len(results) >= 1:
         raise exceptions.UniquenessError(model=model, conflicting_results=results)
 
+    return True
```
JacksonMaxfield, line 52: Can you update docstring to say:
```
Returns
----------
is_unique: bool
    Boolean representing if the model is unique in the collection.
```

cdp_backend/tests/database/test_validators.py
```diff
@@ -51,6 +50,31 @@ def test_uniqueness_validation(
...
+        # Get model spec from models module
+        spec = getattr(models, model_name)
+
+        # Get instance of model
+        instance = spec.Example()
+
+        # Validate
+        assert expected == validators.model_is_unique(instance)
+
+
```
JacksonMaxfield, line 77: Regardless of if we agree that the function shouldn't raise and error or not these tests should be combined with the other tests for this function imo. They are 99% the same code and looking at it you should be able to just add the `expected` parameter to the above tests and get the same behavior.

If we remove the exception from the function then this becomes even easier to test.

cdp_backend/database/validators.py
isaacna: Yeah I was actually thinking that too, but wasn't sure if you had a specific reason for raising error or not. I'd prefer returning True/False rather than having to catch an handle a potential exception

cdp_backend/database/validators.py
isaacna: Will do

cdp_backend/tests/database/test_validators.py
isaacna: Sounds good, I did this bc I was having issues with `expected` param when one was a `bool` and the other output was an exception. But yeah I'll change the output to True/False and make it into one test

isaacna:
Looks like code coverage dropped bc we aren't using UniquenessError anymore. I can delete that file or we can keep it in case it may be used in the future

cdp_backend/tests/database/test_validators.py
```diff
@@ -49,6 +39,7 @@ def test_uniqueness_validation(
 
         # Validate
         validators.model_is_unique(instance)
+        assert expected == validators.model_is_unique(instance)
```
JacksonMaxfield, line 42: Can you remove the extra `validators.model_is_unique` call?

JacksonMaxfield (commented):
Feel free to delete the now unused exception.

cdp_backend/tests/database/test_validators.py
isaacna: Oops yeah my b

JacksonMaxfield (approved):
Nice. Thanks for all the changes! This is perfect. Feel free to merge whenever.

Issue #16, opened 2020-11-14
Add EventProcessingPipeline


JacksonMaxfield:
## Use Case
*Please provide a use case to help us understand your request in context*

Add the core pipeline to the repo.

## Solution
*Please describe your ideal solution*

Bring over the major components of the existing `cdptools.pipelines.EventGatherPipeline` but as a Prefect `Flow`. All files should be handled with `gcsfs` / `fsspec`.

## Alternatives
*Please describe any alternatives you've considered, even if you've dismissed them*

JacksonMaxfield:
Add more documentation:

This relies on #10 as the input to this entire pipeline is the output of some user defined `get_events` function.

i.e.
```python
# in the user script
@task
def get_events() -> List[EventData]:
    ...

# in our pipeline
from user_script import get_events

with Flow("event-gather-pipeline") as flow:
    events = get_events()
    ...
```

JacksonMaxfield:
@Shak2000 when you wrap up on the thumbnail pipeline addition feel free to add in your PR:

"Resolves #16"

JacksonMaxfield:
Resolved in #78

PR #32, opened 2020-12-19
feature/model-validator-fix-and-update

isaacna:
* Fixed query building bug when primary key is another Model/ReferenceField

* Create class for validation output and add conflicting model results

 - Relevant to #16 
 - Query was erroring out in `is_model_unique` when we were filtering by a `ReferenceField` (when querying something like `Event` which has `Body` as a primary key.
 - Fixed how we build the query because the way that the queries are built for `fireo` [is different when you want to filter by a `ReferenceField`](https://octabyte.io/FireO/querying-data/querying-and-filtering/#filter-with-reference-model), and this was only [recently added in 1.3.7](https://github.com/octabytes/FireO/releases/tag/v1.3.7). 
 - Added `UniquenessValidation` object because in the event gather pipeline we want to retrieve the existing db model(s) if it exists (Ex. if `Body` exists, we want to use that model as a `ReferenceField` in the `Event`). This way we can properly build the db models bottom-up from the ingestion models without having to make extra queries to the database.
 - Added test case for `Event` since it has a `ReferenceField` primary key
 - Manually tested that this works

Thanks for contributing!

### DIFF
--- a/cdp_backend/database/validators.py
+++ b/cdp_backend/database/validators.py
@@ -29,8 +29,8 @@ class UniquenessValidation:
     """
     An object containing uniqueness data of a database model object.
 
-    Notes
-    -----
+    Parameters
+    ----------
     is_unique is a boolean on whether the model is unique by primary key
     in the database collection.
 
@@ -42,7 +42,7 @@ class UniquenessValidation:
     conflicting_models: List[Model]
 
 
-def model_is_unique(model: Model) -> UniquenessValidation:
+def get_model_uniqueness(model: Model) -> UniquenessValidation:
     """
     Validate that the primary keys of a to-be-uploaded model are unique
     when compared to the collection.
@@ -60,10 +60,11 @@ def model_is_unique(model: Model) -> UniquenessValidation:
     Examples
     --------
     >>> from cdp_backend.database import models
-    ... from cdp_backend.database.validators import model_is_unique
+    ... from cdp_backend.database.validators import get_model_uniqueness
     ...
     ... b = models.Body.Example()
-    ... if model_is_unique(b):
+    ... b_uniqueness = get_model_uniqueness(b)
+    ... if b_uniqueness.is_unique:
     ...     b.save()
     """
     # Initialize query
--- a/cdp_backend/database/validators.py
+++ b/cdp_backend/database/validators.py
@@ -3,7 +3,8 @@
 
 import logging
 import re
-from typing import Optional
+from typing import List, Optional
+from dataclasses import dataclass
 
 from fireo.models import Model
 from fsspec.core import url_to_fs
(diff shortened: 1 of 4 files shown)


cdp_backend/database/validators.py
```diff
@@ -23,10 +24,28 @@
...
+    An object containing uniqueness data of a database model object.
+
+    Notes
+    -----
+    is_unique is a boolean on whether the model is unique by primary key
+    in the database collection.
+
+    conflicting_models is a List[Model] of all existing models that share
+    the same primary keys as the input model.
+    """
```
JacksonMaxfield, line 39: Should just be `Parameters` in docstring no?

JacksonMaxfield (approved):
This is such a good catch!

Specifically related to this PR, can you version bump `fireo` in dependencies to be `~=1.3.7` to make sure it will always install correctly.

Regarding pulumi preview. I know it's not related to this PR but could you as a part of this PR just remove [this section](https://github.com/CouncilDataProject/cdp-backend/blob/main/.github/workflows/check-pr.yml#L86) of the check-pr workflow.

This should fix the PR checks. It's failing because your fork of the library doesn't have the pulumi credentials.

cdp_backend/database/validators.py
```diff
@@ -23,10 +24,28 @@
...
+
+    is_unique: bool
+    conflicting_models: List[Model]
+
+
+def model_is_unique(model: Model) -> UniquenessValidation:
```
JacksonMaxfield, line 45: I would change this function name to `get_model_uniqueness` or similar. The `*_is_*` naming pattern is specific to when a boolean is returned.

Related, you will need to update this docstring to say:
```python
"""
Examples
--------
>>> from cdp_backend.database import models
... from cdp_backend.database.validators import get_model_uniqueness
...
... b = models.Body.Example()
... b_unique = get_model_uniqueness(b)
... if b_unique.is_unique:
...     b.save()
"""
```

JacksonMaxfield (changes requested):

cdp_backend/database/validators.py
isaacna: Good catch I missed that

cdp_backend/database/validators.py
```diff
@@ -23,10 +24,28 @@
...
+    An object containing uniqueness data of a database model object.
+
+    Parameters
+    ----------
+    is_unique is a boolean on whether the model is unique by primary key
+    in the database collection.
+
+    conflicting_models is a List[Model] of all existing models that share
+    the same primary keys as the input model.
+    """
```
JacksonMaxfield, line 39: Last change is to put the parameters into the normal format:
```python
"""
Parameters
----------
is_unique: bool
    ...

conflicting_models: List[Model]
    ...
"""
```

JacksonMaxfield (approved):
Once the `Parameters` docstring is fixed feel free to merge whenever.

cdp_backend/database/validators.py
isaacna: oops missed that, fixed

PR #33, opened 2020-12-21
feature/event-gather-pipeline-minimal-db-upload

isaacna:
 - Incremental work of #16 
 - Started the event gather pipeline but doing it incrementally for feedback to avoid too large of a PR 
 - This only covers the "upload to database" part of the pipeline, and only for a minimal event (body and sessions). I plan on having separate PR's for the audio/transcript creation and marked with TODO's accordingly
 - Used `example_get_events_func` to make testing easier and as a placeholder, but I'm assuming we want to pass the callable function some other way for different deployments later. Is there an easy way to do this via command line args?
 - I thought it would be pretty complicated/not necessary to write end-to-end tests with all the interactions that don't necessarily return anything at the end, so I plan on writing tests mostly at the task level, but lmk if you think otherwise
 - manually tested this with my personal firestore app

Thanks for contributing!

### DIFF
--- a/cdp_backend/bin/cdp_event_gather.py
+++ b/cdp_backend/bin/cdp_event_gather.py
@@ -5,7 +5,6 @@ import argparse
 import logging
 import sys
 import traceback
-
 from importlib import import_module
 from pathlib import Path
 from typing import Callable
@@ -35,10 +34,7 @@ class Args(argparse.Namespace):
         p.add_argument(
             "-g",
             "--google-credentials-file",
-            default=(
-                Path(__file__).parent.parent.parent
-                / "stg-cdp-seattle-a910f-45df66e72d75.json"
-            ),
+            default=(Path(__file__).parent.parent.parent / "cdp-dev-infra-creds.json"),
             type=Path,
             dest="google_credentials_file",
             help="Path to the Google Service Account Credentials JSON file.",
--- a/cdp_backend/bin/cdp_event_gather.py
+++ b/cdp_backend/bin/cdp_event_gather.py
@@ -10,7 +10,7 @@ from importlib import import_module
 from pathlib import Path
 from typing import Callable
 
-import cdp_backend.pipeline.cdp_event_gather_pipeline as pipeline
+from ..pipeline import cdp_event_gather_pipeline as pipeline
 
 ###############################################################################
 
--- a/cdp_backend/bin/cdp_event_gather.py
+++ b/cdp_backend/bin/cdp_event_gather.py
@@ -8,7 +8,7 @@ import sys
 import traceback
 from datetime import datetime
 
-import cdp_backend.pipeline.event_gather_pipeline as pipeline
+import cdp_backend.pipeline.cdp_event_gather_pipeline as pipeline
 from cdp_backend.pipeline.ingestion_models import (
     EXAMPLE_MINIMAL_EVENT,
     EventIngestionModel,
@@ -33,8 +33,8 @@ class Args(argparse.Namespace):
 
     def __parse(self) -> None:
         p = argparse.ArgumentParser(
-            prog="event_gather",
-            description="Gather events through running the pipeline.",
+            prog="cdp_event_gather",
+            description="Gather, process, and store event data to CDP infrastructure.",
         )
         p.add_argument(
             "-g",
@@ -45,7 +45,7 @@ class Args(argparse.Namespace):
             ),
             type=Path,
             dest="google_credentials_file",
-            help="Path to the google credentials file.",
+            help="Path to the Google Service Account Credentials JSON file.",
         )
         p.parse_args(namespace=self)
 
@@ -62,7 +62,7 @@ def main() -> None:
 
         credentials_file = args.google_credentials_file
 
-        flow = pipeline.create_event_gather_flow(
+        flow = pipeline.create_cdp_event_gather_flow(
             example_get_events_func, credentials_file
         )
 
--- a/cdp_backend/bin/cdp_event_gather.py
+++ b/cdp_backend/bin/cdp_event_gather.py
@@ -3,18 +3,14 @@
 
 import argparse
 import logging
-from pathlib import Path
 import sys
 import traceback
-from datetime import datetime
+
+from importlib import import_module
+from pathlib import Path
+from typing import Callable
 
 import cdp_backend.pipeline.cdp_event_gather_pipeline as pipeline
-from cdp_backend.pipeline.ingestion_models import (
-    EXAMPLE_MINIMAL_EVENT,
-    EventIngestionModel,
-)
-
-from typing import List
 
 ###############################################################################
 
@@ -47,13 +43,24 @@ class Args(argparse.Namespace):
             dest="google_credentials_file",
             help="Path to the Google Service Account Credentials JSON file.",
         )
+        p.add_argument(
+            "-e",
+            "--get_events_function_path",
+            type=Path,
+            dest="get_events_function_path",
+            help=(
+                "Path to the function (including function name) that "
+                "supplies event data to the CDP event gather pipeline."
+            ),
+        )
         p.parse_args(namespace=self)
 
 
-def example_get_events_func() -> List[EventIngestionModel]:
-    event = EXAMPLE_MINIMAL_EVENT
-    event.sessions[0].session_datetime = datetime(2019, 4, 13)
-    return [event]
+def import_get_events_func(func_path: Path) -> Callable:
+    path, func_name = str(func_path).rsplit(".", 1)
+    mod = import_module(path)
+
+    return getattr(mod, func_name)
 
 
 def main() -> None:
@@ -62,9 +69,9 @@ def main() -> None:
 
         credentials_file = args.google_credentials_file
 
-        flow = pipeline.create_cdp_event_gather_flow(
-            example_get_events_func, credentials_file
-        )
+        get_events_func = import_get_events_func(args.get_events_function_path)
+
+        flow = pipeline.create_cdp_event_gather_flow(get_events_func, credentials_file)
 
         # TODO make the flow run on a certain type of Executor?
         flow.run()
--- a/cdp_backend/bin/cdp_event_gather.py
+++ b/cdp_backend/bin/cdp_event_gather.py
@@ -73,7 +73,6 @@ def main() -> None:
 
         flow = pipeline.create_cdp_event_gather_flow(get_events_func, credentials_file)
 
-        # TODO make the flow run on a certain type of Executor?
         flow.run()
 
     except Exception as e:
--- a/cdp_backend/bin/create_ingestion_model_doc.py
+++ b/cdp_backend/bin/create_ingestion_model_doc.py
@@ -82,10 +82,8 @@ def _construct_ingestion_model_doc(template_file: Path, output_file: Path) -> Pa
 
     # Fill the template with values
     filled = template.render(
-        minimal_event_data=pformat(  # type: ignore
-            minimal_event_data, sort_dicts=False
-        ),
-        filled_event_data=pformat(filled_event_data, sort_dicts=False),  # type: ignore
+        minimal_event_data=pformat(minimal_event_data, sort_dicts=False),
+        filled_event_data=pformat(filled_event_data, sort_dicts=False),
     )
 
     # Store filled template
--- a/cdp_backend/bin/create_ingestion_model_doc.py
+++ b/cdp_backend/bin/create_ingestion_model_doc.py
@@ -82,8 +82,10 @@ def _construct_ingestion_model_doc(template_file: Path, output_file: Path) -> Pa
 
     # Fill the template with values
     filled = template.render(
-        minimal_event_data=pformat(minimal_event_data, sort_dicts=False),
-        filled_event_data=pformat(filled_event_data, sort_dicts=False),
+        minimal_event_data=pformat(  # type: ignore
+            minimal_event_data, sort_dicts=False
+        ),
+        filled_event_data=pformat(filled_event_data, sort_dicts=False),  # type: ignore
     )
 
     # Store filled template
--- a/cdp_backend/bin/event_gather.py
+++ b/cdp_backend/bin/event_gather.py
@@ -66,9 +66,7 @@ def main() -> None:
             example_get_events_func, credentials_file
         )
 
-        # Run
-        # flow.run(executor=DaskExecutor(address=cluster.scheduler_address))
-        # flow.register("CDP Pipelines")
+        # TODO make the flow run on a certain type of Executor?
         flow.run()
 
     except Exception as e:
--- a/cdp_backend/bin/event_gather.py
+++ b/cdp_backend/bin/event_gather.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import argparse
+import logging
+from pathlib import Path
+import sys
+import traceback
+from datetime import datetime
+
+import cdp_backend.pipeline.event_gather_pipeline as pipeline
+from cdp_backend.pipeline.ingestion_models import (
+    EXAMPLE_MINIMAL_EVENT,
+    EventIngestionModel,
+)
+
+from typing import List
+
+###############################################################################
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="[%(levelname)4s: %(module)s:%(lineno)4s %(asctime)s] %(message)s",
+)
+log = logging.getLogger(__name__)
+
+###############################################################################
+
+
+class Args(argparse.Namespace):
+    def __init__(self) -> None:
+        self.__parse()
+
+    def __parse(self) -> None:
+        p = argparse.ArgumentParser(
+            prog="event_gather",
+            description="Gather events through running the pipeline.",
+        )
+        p.add_argument(
+            "-g",
+            "--google-credentials-file",
+            default=(
+                Path(__file__).parent.parent.parent
+                / "stg-cdp-seattle-a910f-45df66e72d75.json"
+            ),
+            type=Path,
+            dest="google_credentials_file",
+            help="Path to the google credentials file.",
+        )
+        p.parse_args(namespace=self)
+
+
+def example_get_events_func() -> List[EventIngestionModel]:
+    event = EXAMPLE_MINIMAL_EVENT
+    event.sessions[0].session_datetime = datetime(2019, 4, 13)
+    return [event]
+
+
+def main() -> None:
+    try:
+        args = Args()
+
+        credentials_file = args.google_credentials_file
+
+        flow = pipeline.create_event_gather_flow(
+            example_get_events_func, credentials_file
+        )
+
+        # Run
+        # flow.run(executor=DaskExecutor(address=cluster.scheduler_address))
+        # flow.register("CDP Pipelines")
+        flow.run()
+
+    except Exception as e:
+        log.error("=============================================")
+        log.error("\n\n" + traceback.format_exc())
+        log.error("=============================================")
+        log.error("\n\n" + str(e) + "\n")
+        log.error("=============================================")
+        sys.exit(1)
+
+
+###############################################################################
+# Allow caller to directly run this module (usually in development scenarios)
+
+if __name__ == "__main__":
+    main()
--- a/cdp_backend/pipeline/cdp_event_gather_pipeline.py
+++ b/cdp_backend/pipeline/cdp_event_gather_pipeline.py
@@ -12,6 +12,7 @@ from cdp_backend.database.validators import get_model_uniqueness
 from prefect import Flow, task
 
 import fireo
+from fireo.models import Model
 
 ###############################################################################
 
@@ -36,57 +37,40 @@ def create_cdp_event_gather_flow(
         events: List[EventIngestionModel] = get_events_func()
 
         for event in events:
-            process_event(event)
+            # TODO create/get transcript
+            # TODO create/get audio (happens as part of transcript process)
+
+            # Upload calls for minimal event
+            body_ref = upload_db_model(create_body_from_ingestion_model(event.body))
+
+            # TODO add upload calls for non-minimal event
+
+            event_ref = upload_db_model(
+                create_event_from_ingestion_model(event, body_ref)
+            )
+
+            for session in event.sessions:
+                # upload_session(session, event_ref)
+                upload_db_model(create_session_from_ingestion_model(session, event_ref))
 
     return flow
 
 
-def process_event(event: EventIngestionModel) -> None:
-    # TODO create/get transcript
-    # TODO create/get audio (happens as part of transcript process)
-
-    # Upload converted ingestion models to databae
-    upload_to_database(event)
-
-
-def upload_to_database(event: EventIngestionModel) -> None:
-    # Upload calls for minimal event
-    body_ref = upload_body(event.body)
-
-    # TODO add upload calls for non-minimal event
-
-    event_ref = upload_event(event, body_ref)
-
-    for session in event.sessions:
-        upload_session(session, event_ref)
-
-
 @task
-def upload_event(
-    event: EventIngestionModel, body_ref: db_models.Body
-) -> db_models.Event:
-    db_event = db_models.Event()
+def upload_db_model(model: Model) -> Model:
+    uniqueness_validation = get_model_uniqueness(model)
 
-    # Required fields
-    db_event.body_ref = body_ref
-
-    # Assume that session is same day as event
-    db_event.event_datetime = event.sessions[0].session_datetime
-
-    # TODO add optional fields
-
-    uniqueness_validation = get_model_uniqueness(db_event)
     if uniqueness_validation.is_unique:
-        db_event.save()
-        log.info(f"Saved new Event with document id={db_event.id}.")
+        model.save()
+        log.info(f"Saved new {model.__class__.__name__} with document id={model.id}.")
     else:
         return uniqueness_validation.conflicting_models[0]
 
-    return db_event
+    return model
 
 
 @task
-def upload_body(body: Body) -> db_models.Body:
+def create_body_from_ingestion_model(body: Body) -> db_models.Body:
     db_body = db_models.Body()
 
     # Required fields
@@ -107,18 +91,30 @@ def upload_body(body: Body) -> db_models.Body:
     if body.external_source_id:
         db_body.external_source_id = body.external_source_id
 
-    uniqueness_validation = get_model_uniqueness(db_body)
-    if uniqueness_validation.is_unique:
-        db_body.save()
-        log.info(f"Saved new Body with document id={db_body.id}.")
-    else:
-        return uniqueness_validation.conflicting_models[0]
-
     return db_body
 
 
 @task
-def upload_session(session: Session, event_ref: db_models.Event) -> db_models.Session:
+def create_event_from_ingestion_model(
+    event: EventIngestionModel, body_ref: db_models.Body
+) -> db_models.Event:
+    db_event = db_models.Event()
+
+    # Required fields
+    db_event.body_ref = body_ref
+
+    # Assume that session is same day as event
+    db_event.event_datetime = event.sessions[0].session_datetime
+
+    # TODO add optional fields
+
+    return db_event
+
+
+@task
+def create_session_from_ingestion_model(
+    session: Session, event_ref: db_models.Event
+) -> db_models.Session:
     db_session = db_models.Session()
 
     # Required fields
@@ -138,12 +134,5 @@ def upload_session(session: Session, event_ref: db_models.Event) -> db_models.Se
 
     if session.external_source_id:
         db_session.external_source_id = session.external_source_id
-
-    uniqueness_validation = get_model_uniqueness(db_session)
-    if uniqueness_validation.is_unique:
-        db_session.save()
-        log.info(f"Saved new Session with document id={db_session.id}.")
-    else:
-        return uniqueness_validation.conflicting_models[0]
 
     return db_session
--- a/cdp_backend/pipeline/cdp_event_gather_pipeline.py
+++ b/cdp_backend/pipeline/cdp_event_gather_pipeline.py
@@ -3,11 +3,12 @@
 
 from datetime import datetime
 import logging
-from typing import Callable, List
+from typing import Any, Callable, List
 
-from cdp_backend.pipeline.ingestion_models import EventIngestionModel, Body, Session
-import cdp_backend.database.models as db_models
-from cdp_backend.database.validators import get_model_uniqueness
+from .ingestion_models import EventIngestionModel, Body, Session
+from ..database import models as db_models
+from ..database import exceptions
+from ..database.validators import get_model_uniqueness
 
 from prefect import Flow, task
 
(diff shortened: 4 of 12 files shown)


cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -0,0 +1,146 @@
+from datetime import datetime
```
isaacna, line 1: Forgot:
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
```
I'll add that on the next commit

cdp_backend/bin/event_gather.py
```diff
@@ -0,0 +1,85 @@
...
+###############################################################################
+
+
+class Args(argparse.Namespace):
+    def __init__(self) -> None:
+        self.__parse()
+
+    def __parse(self) -> None:
+        p = argparse.ArgumentParser(
+            prog="event_gather",
```
JacksonMaxfield, line 36: Prefix with `cdp` -> `cdp_event_gather`.

cdp_backend/bin/event_gather.py
```diff
@@ -0,0 +1,85 @@
...
+            description="Gather events through running the pipeline.",
```
JacksonMaxfield, line 37: I would change to `"Gather, process, and store event data to CDP infrastructure."`

cdp_backend/bin/event_gather.py
```diff
@@ -0,0 +1,85 @@
...
+        p.add_argument(
+            "-g",
+            "--google-credentials-file",
+            default=(
+                Path(__file__).parent.parent.parent
+                / "stg-cdp-seattle-a910f-45df66e72d75.json"
+            ),
+            type=Path,
+            dest="google_credentials_file",
+            help="Path to the google credentials file.",
```
JacksonMaxfield, line 48: Can you add that this is specifically a "Google Service Account Credentials JSON"

cdp_backend/bin/event_gather.py
```diff
@@ -0,0 +1,85 @@
...
+class Args(argparse.Namespace):
+    def __init__(self) -> None:
+        self.__parse()
+
+    def __parse(self) -> None:
+        p = argparse.ArgumentParser(
+            prog="event_gather",
+            description="Gather events through running the pipeline.",
+        )
+        p.add_argument(
```
JacksonMaxfield, line 39: Need to add an argument that is "path to `get_events` function file"

cdp_backend/bin/event_gather.py
```diff
@@ -0,0 +1,85 @@
...
+                / "stg-cdp-seattle-a910f-45df66e72d75.json"
+            ),
+            type=Path,
+            dest="google_credentials_file",
+            help="Path to the google credentials file.",
+        )
+        p.parse_args(namespace=self)
+
+
+def example_get_events_func() -> List[EventIngestionModel]:
```
JacksonMaxfield, line 53: This function should live in a separate file and should be read in using some sort of importlib + filepath  + finding `get_events`

cdp_backend/bin/event_gather.py
```diff
@@ -0,0 +1,85 @@
...
+        args = Args()
+
+        credentials_file = args.google_credentials_file
+
+        flow = pipeline.create_event_gather_flow(
+            example_get_events_func, credentials_file
+        )
+
+        # TODO make the flow run on a certain type of Executor?
+        flow.run()
```
JacksonMaxfield, line 70: Still debating. I think maybe this is determined with arguments. I.e. default it runs synchronously (like current) but on backfill jobs it runs w/ dask or something? Don't change anything for now just thinking about it.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -0,0 +1,146 @@
...
+def create_event_gather_flow(
+    get_events_func: Callable,
+    credentials_file: str,
+) -> Flow:
+    # Initialize fireo connection
+    fireo.connection(from_file=credentials_file)
+
+    # Create flow
+    with Flow("Event Gather Pipeline") as flow:
+        events: List[EventIngestionModel] = get_events_func()
```
JacksonMaxfield, line 33: Woah, didn't know you could type returns like that. I assume `mypy` enforced this? I personally like it.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -0,0 +1,146 @@
...
+
+        for event in events:
+            process_event(event)
```
JacksonMaxfield, line 36: We could `map` this but we commonly run into race conditions with `ThreadPoolExecutor` as it is now so going to say "sadly, for loop is the way to go"

JacksonMaxfield: That said. I think this `process_event` -> `upload_to_database` -> `upload_body` -> etc. is a bit odd.

I would have all of the final children tasks live in this loop.

I.E.
```python
for event in events:
    body_ref = upload_body(event.body)
    event_ref = upload_event(event, body_ref)
    for session in event.sessions:
        upload_session(session)
```

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -0,0 +1,146 @@
...
+
+    return flow
+
+
+def process_event(event: EventIngestionModel) -> None:
+    # TODO create/get transcript
+    # TODO create/get audio (happens as part of transcript process)
+
+    # Upload converted ingestion models to databae
+    upload_to_database(event)
```
JacksonMaxfield, line 46: These todos should be their own `@task` declarations imo.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def upload_to_database(event: EventIngestionModel) -> None:
+    # Upload calls for minimal event
+    body_ref = upload_body(event.body)
+
+    # TODO add upload calls for non-minimal event
+
+    event_ref = upload_event(event, body_ref)
```
JacksonMaxfield, line 55: There is a part of me that wants to split out some of the logic contained in these `upload_*` tasks. I.E. separation of getting the first / earliest datetime session and applying that datetime to the event itself. Thoughts?

My reasoning would be easier testing so that you don't need to mock database interactions on simply business logic.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # Optional fields
+    if body.end_datetime:
+        db_body.end_datetime = body.end_datetime
+
+    if body.description:
+        db_body.description = body.description
+
+    if body.external_source_id:
+        db_body.external_source_id = body.external_source_id
```
JacksonMaxfield, line 105: This is a perfect example of my above comment about separation of business logic vs upload management. This could be tested really easily without mocking uploads imo.

JacksonMaxfield: Maybe make these functions be `create_body_from_ingestion_model` and `create_event_from_ingestion_model`, etc.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -0,0 +1,146 @@
...
+
+    uniqueness_validation = get_model_uniqueness(db_body)
+    if uniqueness_validation.is_unique:
+        db_body.save()
+        log.info(f"Saved new Body with document id={db_body.id}.")
+    else:
+        return uniqueness_validation.conflicting_models[0]
```
JacksonMaxfield, line 112: This then further frees up this code to be a general `handle_model_upload` task that tasks some model and handles the exception because I see you are duplicating this code a bit.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -0,0 +1,146 @@
...
+
+    if session.external_source_id:
+        db_session.external_source_id = session.external_source_id
+
+    uniqueness_validation = get_model_uniqueness(db_session)
+    if uniqueness_validation.is_unique:
+        db_session.save()
+        log.info(f"Saved new Session with document id={db_session.id}.")
+    else:
+        return uniqueness_validation.conflicting_models[0]
```
JacksonMaxfield, line 144: Same business logic and upload operation separation comments here.

cdp_backend/tests/pipeline/test_event_gather_pipeline.py
```diff
@@ -0,0 +1,53 @@
...
+    with mock.patch(
+        "cdp_backend.pipeline.event_gather_pipeline.get_model_uniqueness"
+    ) as mock_uniqueness_validator:
+        mock_uniqueness_validator.return_value = UniquenessValidation(
+            False, [db_body_example]
+        )
+
+        db_body = pipeline.upload_body.run(ingestion_body)
+        assert db_body.name == ingestion_body.name
+        assert db_body.is_active == ingestion_body.is_active
```
JacksonMaxfield, line 53: After changes come in for business logic separation you can add tests for all the optional + required parameter mutations and then a single one for model upload.

cdp_backend/tests/pipeline/test_event_gather_pipeline.py
```diff
@@ -0,0 +1,53 @@
...
+ingestion_body = ingestion_models.Body(
+    name=db_body_example.name, is_active=db_body_example.is_active
+)
+
+
+def test_create_event_gather_flow() -> None:
+    with mock.patch("fireo.connection") as mock_connector:
+        mock_connector.return_value = None
+        flow = pipeline.create_event_gather_flow(example_get_events_func, "/fake/path")
+        assert isinstance(flow, Flow)
```
JacksonMaxfield, line 24: Note to self: Create a bin script to generate the flow visualization to attach to documentation after this / other PRs merge like the current doc generation scripts.

setup.py
```diff
@@ -50,6 +50,7 @@
     "pandas~=1.1.3",
     "pulumi~=2.12.1",
     "pulumi-gcp~=4.1.0",
+    "prefect==0.13.0"
```
JacksonMaxfield, line 53: They just released a new 0.14.0 version is shouldn't affect this PR but would be nice to use if possible.

Can we safely change this to `prefect~=0.14.0`

setup.py
```diff
@@ -84,6 +85,7 @@
                 "create_cdp_ingestion_model_doc="
                 "cdp_backend.bin.create_ingestion_model_doc:main"
             ),
+            "event_gather=cdp_backend.bin.event_gather:main",
```
JacksonMaxfield, line 88: Related to first comment: prefix with `cdp` -> `cdp_event_gather`

JacksonMaxfield (changes requested):
Overall, looking real solid. I love how simple the fireo stuff becomes. And the business logic generally makes sense so far!

cdp_backend/pipeline/event_gather_pipeline.py
tohuynh: Wouldn't this make `processs_event` a flow? `@task` are discrete actions of a flow right?

cdp_backend/pipeline/event_gather_pipeline.py
tohuynh: I agree with Jackson. That's a good idea.

cdp_backend/pipeline/event_gather_pipeline.py
tohuynh: Another note: create/get transcript and create/get audio todos are for sessions. I think we'd still need a `a process_session`.

JacksonMaxfield: I believe that will come in the "next PR"

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: Yes each @task is a discrete step in the workflow which is also why I want to bring them to the top level.

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Yeah I find doing this pretty helpful just for writing code itself, wouldn't mind making it standard practice

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Maybe my understanding of prefect was wrong, but I thought since the the tasks in each iteration of the for loop are independent of each other, prefect "magic" would handle that properly when generating the DAG and run tasks in parallel if we chose an executor like dask. 

Also it looks like prefect has some mapping examples that I think we can do instead of the for loop: https://docs.prefect.io/core/concepts/mapping.html#prefect-approach

Also, are there race conditions across mulitple event processes that I'm missing? Do you men the potential of 2 events modifying a shared document like `Body` or something?

isaacna: And yeah I was planning on handling the transcript/audio at a session level in the next PR, but put it in the process_event comments for now.

Yeah I may have over modularized the upload code, I can simplify that a bit

JacksonMaxfield: You are correct in that you are constructing a DAG here and that yes prefect "magic" will unpack the for loop for us! My bad in my thinking!

But I wouldn't use a non-synchronous executor for this specific workflow unless the user is backfilling a lot of events because exactly like you mentioned there exists race conditions not common for `Body` but very common for minutes items. Currently (in `cdptools`) we treat each session as it's own event and as such if two events are being uploaded in parrallel it's pretty common to have a race condition for `MinutesItem` or `Matter` documents.

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Are we trying to make each instance of processing a single event a flow? The way I was approaching it was that the flow is the multiple instances of processing an event, and a single `process_event` call is not a flow itself. `process_event` is just a container method (for readability) and sets the tasks for each event we are trying to process, and then we rely on prefect to determine the workflow and what tasks are dependent on each other. 

Does bringing the tasks to the top level change the functionality at all or is it just a prefect coding standard thing?

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Yeah good point, I think that's the right call and will make the testing much easier

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: No don't make `process_event` it's own flow. You had the right idea here in that this is all just DAG definition I just don't like the styling of having a container of prefect tasks is all. I think the flow defintion should have everything in it without any "multi-task containers" basically.

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: I can see this being useful for specifically "functions or callables imported from some other library" but if the function is defined in the same file I think it may be overkill. But yea. definitely nice.

isaacna:
Not sure why the infrastructure tests are still failing.

It looks like the mypy checks are also failing due to how I am unit testing the tasks. I think I know what the issue is but I've asked a question in the prefect community just to make sure: https://github.com/PrefectHQ/prefect/discussions/3896

isaacna:
I ignored the mypy tests that were failing on some of the prefect lines. I think I know what the issue but I brought it up here too: https://github.com/PrefectHQ/prefect/discussions/3896

I still don't fully understand why the pulumi test was failing, but from looking into it a bit I think it was because `@pulumi.runtime.test` actually initializes an asyncio event loop to run the test in, and for some reason importing prefect in one of the later files like `test_cdp_event_gather_pipeline` was causing the the asynchronous tasks that pulumi generates internally to be attached to a different event loop. Importing prefect into the `infrastructure/__init__.py` fixed this for me

cdp_backend/pipeline/cdp_event_gather_pipeline.py
```diff
@@ -0,0 +1,138 @@
...
+
+@task
+def upload_db_model(model: Model) -> Model:
+    uniqueness_validation = get_model_uniqueness(model)
+
+    if uniqueness_validation.is_unique:
+        model.save()
+        log.info(f"Saved new {model.__class__.__name__} with document id={model.id}.")
+    else:
+        return uniqueness_validation.conflicting_models[0]
```
JacksonMaxfield, line 67: There is an interesting decision to make here. I think I said in documentation somewhere that if the document already exists the db, and the non-primary fields are different from how they are currently stored, it would update the fields that aren't primary keys. (Change in body description for example). Not quite sure how to do that.... I know each model definition has the primary keys labeled so it would be possible to check that the found model has different non-primary field values and then update and resave?

Idk, thoughts here would be nice. @tohuynh too.

cdp_backend/pipeline/cdp_event_gather_pipeline.py
```diff
@@ -0,0 +1,138 @@
...
+def create_event_from_ingestion_model(
+    event: EventIngestionModel, body_ref: db_models.Body
+) -> db_models.Event:
+    db_event = db_models.Event()
+
+    # Required fields
+    db_event.body_ref = body_ref
+
+    # Assume that session is same day as event
+    db_event.event_datetime = event.sessions[0].session_datetime
```
JacksonMaxfield, line 107: Because there can be a session index on each session model, and the zeroth session isn't guaranteed to be the datetime earliest, can you do a list comprehension to find the earliest one and use that? (I don't think we would ever have a multi-day event but :shrug: never know)

cdp_backend/pipeline/cdp_event_gather_pipeline.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Required fields
+    db_session.event_ref = event_ref
+    db_session.session_datetime = session.session_datetime
+    db_session.video_uri = session.video_uri
+
+    if session.session_index:
+        db_session.session_index = session.session_index
+    else:
+        # Is this how we want to handle when session_index isn't provided?
```
JacksonMaxfield, line 128: No, if no session index isn't provided it should be the index of the session in the parent list. Which means you will probably need to do a pre-process step on the whole session list prior to this task.

setup.py
```diff
@@ -50,6 +50,7 @@
     "pandas~=1.1.3",
     "pulumi~=2.12.1",
     "pulumi-gcp~=4.1.0",
+    "prefect==0.14.0"
```
JacksonMaxfield, line 53: Make this specifically `~=` instead of `==`.

cdp_backend/tests/pipeline/test_cdp_event_gather_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+    fields = [attr for attr in dir(ingestion_model) if not attr.startswith("__")]
+
+    for field in fields:
+        ingestion_value = getattr(ingestion_model, field)
+
+        # Minimal models may be missing some values
+        # Some fields like reference fields don't match between ingestion and db models
+        # Those are asserted in the more specific methods
+        if ingestion_value and hasattr(expected_db_model, field):
+            assert getattr(expected_db_model, field) == getattr(actual_db_model, field)
```
JacksonMaxfield, line 173: Can you move these to be near the top of the file or at least above where they are used in the tests, was a bit confused as to where they came from haha.

JacksonMaxfield (changes requested):
This is looking much better!

More minor changes and some discussions.

JacksonMaxfield:
> I ignored the mypy tests that were failing on some of the prefect lines. I think I know why this happens but I brought it up here too: [PrefectHQ/prefect#3896](https://github.com/PrefectHQ/prefect/discussions/3896)
> 
> I still don't fully understand why the pulumi test was failing, but from looking into it a bit I think it was because `@pulumi.runtime.test` actually initializes an asyncio event loop to run the test in, and for some reason importing prefect in one of the later files like `test_cdp_event_gather_pipeline` was causing the the asynchronous tasks that pulumi generates internally to be attached to a different event loop. Importing prefect into the `infrastructure/__init__.py` fixed this for me

Yea I have scene some incredible event loop testing due to similar issues. Never with Prefect though. Hopefully someone responds to the discussion post you had. But it's over break -- if no one responds in a couple of days I will post to their Slack.

cdp_backend/pipeline/cdp_event_gather_pipeline.py
isaacna: Sounds good, updating the non-primary fields shouldn't too bad. I think I can just loop through the ingestion model fields and if the found db model differs, update those fields and overwrite.

The other thing I defaulted to was returning the first model, but should we consider what to do if there's more than one? That would be assuming our db has bad data, but that shouldn't happen if this script is the only source of event gathering, or if a race condition appears if we try to parallelize stuff

cdp_backend/pipeline/cdp_event_gather_pipeline.py
isaacna: Works for me, I just made that assumption but can change it just to be safe

cdp_backend/pipeline/cdp_event_gather_pipeline.py
isaacna: > the index of the session in the parent list 

By this do you mean it's actual list index in the parent event's `sessions` list? 

For this ppre-processing logic, in the case that session_index is provided on some of the `session` ingestion models but not the others, do we want to go off the actual `session_index` values or the list indices?

cdp_backend/pipeline/cdp_event_gather_pipeline.py
JacksonMaxfield: I thought about that too. I would say adding a check for if the length of the returned conflicting models list is greater than 1 and raising an error makes sense to me.

cdp_backend/pipeline/cdp_event_gather_pipeline.py
JacksonMaxfield: > By this do you mean it's actual list index in the parent event's `sessions` list?

Yep!

---

Here is how I would do the indexing logic in the mixed index and no index provided case.

If the session index is available / provided, use it. If not, use the list item index. To do so, and this isn't efficient but hey this isn't algorithms class, I would make a list of None's that is of length of `max([len(sessions), max_session_index])` then fill in the indexes that are provided, then replace None in the list with sessions that don't have specific session indexes, then just remove all extra Nones.

---

After typing that out, maybe we just remove the `session_index` from the ingestion model and say that we will create `session_index` in the database solely off of list index...........

cdp_backend/pipeline/cdp_event_gather_pipeline.py
isaacna: Works for me

cdp_backend/pipeline/cdp_event_gather_pipeline.py
isaacna: I'm good with implementing the above logic, I still would preferable to keep `session_index` in the ingestion model in case different implementations on the scraper side make using the order of the list unreliable, like threading over each session

tohuynh: We could always just use the required `datetime` field to infer the sessions' indices. If the list of sessions has one or more missing `session_index`, use the `datetime` field to sort the list and then use the list item indices.

JacksonMaxfield: Sounds good to me! I hadn't thought of using the datetime but that would make sense... at least in the case that no one submits an event with multiple sessions starting at the same datetime.

isaacna: Only thing about sorting by datetime is that at least for [Seattle](http://www.seattlechannel.org/BudgetCommittee), sessions can happen on the same day and from what I can tell the inferre datetime from scraping doesn't get more detailed than day-level.  

So maybe when `session_index` is missing for some, we start a partial list sorted by session_index, and then for the remaining sessions sort by datetime and then append them to that list? 

I think we should prioritize `session_index` first since that's usually explicitly derived from the html. I think the case where the provided `session_index` contradicts the event datetime is pretty unlikely, and in that case that I think that issue should be more on how the scraper is setting that data.

cdp_backend/pipeline/cdp_event_gather_pipeline.py
```diff
@@ -0,0 +1,138 @@
...
+def create_body_from_ingestion_model(body: Body) -> db_models.Body:
+    db_body = db_models.Body()
+
+    # Required fields
+    db_body.name = body.name
+    db_body.is_active = body.is_active
+    if body.start_datetime is None:
+        db_body.start_datetime = datetime.utcnow()
+    else:
+        db_body.start_datetime = body.start_datetime
```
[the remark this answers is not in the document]
isaacna: I'm fine with doing the one liner or keeping them on separate lines. I personally prefer separate lines slightly more out of habit because I'm used to non-Python, but either works for me. @JacksonMaxfield any preference for coding style?

cdp_backend/pipeline/cdp_event_gather_pipeline.py
```diff
@@ -0,0 +1,138 @@
...
+) -> db_models.Session:
+    db_session = db_models.Session()
+
+    # Required fields
+    db_session.event_ref = event_ref
+    db_session.session_datetime = session.session_datetime
+    db_session.video_uri = session.video_uri
+
+    if session.session_index:
+        db_session.session_index = session.session_index
```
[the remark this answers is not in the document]
isaacna: oh that's a really good catch I didn't notice that edge case

cdp_backend/pipeline/cdp_event_gather_pipeline.py
JacksonMaxfield: Yea I think the sort order should be `session_index` first, then `session_datetime`. But also, because I can't remember if we talked about it, did we ever discuss just making `session_index` required in the ingestion model? That would resolve all of these issues imo.

cdp_backend/pipeline/cdp_event_gather_pipeline.py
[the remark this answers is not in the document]
JacksonMaxfield: I personally prefer the elongated form that is currently there. Seen both one liner form and elongated forms many-a-time, and neither one bothers me, so really your call.

cdp_backend/pipeline/cdp_event_gather_pipeline.py
isaacna: I agree, I'll just make `session_index` required. I think this is a better call since this logic should already be determined from the event providing function side imo.

cdp_backend/bin/cdp_event_gather.py
```diff
@@ -0,0 +1,92 @@
...
+    try:
+        args = Args()
+
+        credentials_file = args.google_credentials_file
+
+        get_events_func = import_get_events_func(args.get_events_function_path)
+
+        flow = pipeline.create_cdp_event_gather_flow(get_events_func, credentials_file)
+
+        # TODO make the flow run on a certain type of Executor?
```
JacksonMaxfield, line 76: Can you remove this comment for now. Sorry nitpicky but :shrug:

cdp_backend/bin/create_ingestion_model_doc.py
```diff
@@ -81,8 +81,10 @@ def _construct_ingestion_model_doc(template_file: Path, output_file: Path) -> Pa
 
     # Fill the template with values
     filled = template.render(
-        minimal_event_data=pformat(minimal_event_data, sort_dicts=False),
-        filled_event_data=pformat(filled_event_data, sort_dicts=False),
+        minimal_event_data=pformat(  # type: ignore
+            minimal_event_data, sort_dicts=False
+        ),
+        filled_event_data=pformat(filled_event_data, sort_dicts=False),  # type: ignore
```
JacksonMaxfield, line 87: I am somewhat surprised that `mypy` is complaining here.... can you post the error log if you remove just these type ignores?

isaacna: Yeah, I suddenly started getting the following:
```
cdp_backend/bin/create_ingestion_model_doc.py:84: error: Unexpected keyword argument "sort_dicts" for "pformat"
cdp_backend/bin/create_ingestion_model_doc.py:87: error: Unexpected keyword argument "sort_dicts" for "pformat"
```

I cleaned my project and re-installed all dependencies, so maybe the pprint module got updated?

cdp_backend/bin/cdp_event_gather.py
isaacna: Will do, will remove that in my upcoming set of changes

cdp_backend/bin/create_ingestion_model_doc.py
JacksonMaxfield: Huh. `pprint` is a standard library module so would be surprised if it changed on a patch but I did look up the docs and saw that it didn't have the `sort_dicts` param until py38, so maybe this error occurred in py37?

isaacna: Oh yeah that's probably it, on my machine I recently switched to using py37. I should be good to remove these then since the lint checker runs on py38

isaacna:
Some notes on these changes:
- Added back in the `UniquenessError` to use for cases where num conflicting models > 1. I throw this error on the pipeline side rather than the validator itself so that we can use the conflicting models for something else without throwing error (ex. a script to fix docs with duplicate primary keys)
- Passed in the db key as part of the `update_db_model` method bc when we pass a `Model` in by parameter fireo treats it as a temporary doc and removes it's key/id
- The `assert_db_models_equality` is to check db models for equivalence while ignoring some of the fireo boilerplate stuff

cdp_backend/bin/cdp_event_gather.py
```diff
@@ -0,0 +1,91 @@
...
+            prog="cdp_event_gather",
+            description="Gather, process, and store event data to CDP infrastructure.",
+        )
+        p.add_argument(
+            "-g",
+            "--google-credentials-file",
+            default=(
+                Path(__file__).parent.parent.parent
+                / "stg-cdp-seattle-a910f-45df66e72d75.json"
+            ),
```
JacksonMaxfield, line 41: Can you make this a standard name like `cdp-dev-infra-creds.json`. This is so specific to yours haha.

cdp_backend/pipeline/cdp_event_gather_pipeline.py
```diff
@@ -0,0 +1,184 @@
...
+
+    needs_update = False
+
+    for field in non_primary_db_fields:
+        if hasattr(ingestion_model, field):
+            db_val = getattr(db_model, field)
+            ingestion_val = getattr(ingestion_model, field)
+
+            # If values are different, use the ingestion value
+            # Make sure we don't overwrite with empty values (or should we allow this?)
```
JacksonMaxfield, line 83: I strongly agree that `None` simply means keep the same. So yea, excluding `None` is the right call here.

JacksonMaxfield: Can you add that to the actual ingestion models documentation page actually.

cdp_backend/pipeline/cdp_event_gather_pipeline.py
```diff
@@ -0,0 +1,184 @@
...
+            if db_val != ingestion_val and ingestion_val is not None:
+                setattr(db_model, field, ingestion_val)
+                needs_update = True
+                log.info(f"Updating {db_key} {field} from {db_val} to {ingestion_val}.")
```
JacksonMaxfield, line 87: Beautiful.

cdp_backend/pipeline/cdp_event_gather_pipeline.py
```diff
@@ -0,0 +1,184 @@
...
+            uniqueness_validation.conflicting_models[0].key,
+        )
+
+        return updated_db_model
+    else:
+        raise exceptions.UniquenessError(
+            model=db_model, conflicting_results=uniqueness_validation.conflicting_models
+        )
+
+    return db_model
```
JacksonMaxfield, line 118: Love this new decision flow.

cdp_backend/tests/pipeline/test_cdp_event_gather_pipeline.py
```diff
@@ -0,0 +1,272 @@
...
+    Session,
+)
+
+import copy
+from datetime import datetime
+from fireo.models import Model
+from typing import Any, List
+from prefect import Flow
+from unittest import mock
+import pytest
```
JacksonMaxfield, line 22: Nitpick: I think in my prior comment I asked to make the imports in the pipeline module relative.

In the source files, relative imports are preferred, in the test modules fully qualified are preferred.
So here they should be back to fully qualified.

Idk, Python :shrug: 

Also, can you run:

```bash
pip install isort
isort --recursive
black .
```

This will clean up the import sort order (not the relative vs fully qualified, but just the sort order)
If you use vscode I would drop this into your settings.json as it will make it so that when you save a python file it will auto sort the imports:
```
"[python]": {
        "editor.codeActionsOnSave": {
            "source.organizeImports": true
        }
    }
```

JacksonMaxfield: I have a quick PR with some admin changes that if you merge into will have some isort additions to make it work well with black as well.

JacksonMaxfield (approved):
These changes are great!

Some minor changes / nitpicks to get to and then feel free to merge. I am going to quickly merge in an admin PR to the repo that would awesome if you merged into this branch and made sure everything works.

cdp_backend/tests/pipeline/test_cdp_event_gather_pipeline.py
isaacna: Sounds good, this is a helpful tool!

cdp_backend/bin/cdp_event_gather.py
isaacna: Oh yeah my bad I accidentally left that in to make testing more convenient

cdp_backend/pipeline/cdp_event_gather_pipeline.py
isaacna: Will do

cdp_backend/pipeline/cdp_event_gather_pipeline.py
isaacna: Thanks!

PR #43, opened 2021-01-13
feature/gcsfs-and-fsspec-file-store-functions

isaacna:
 - Part of work leading up to #16, will be used in event gather pipeline during audio creation 
 - Adding file store functions using `gcsfs` and `fsspec` instead of the old filestore class from cdptools
 - Most of the logic surrounding the upload operation remain similar (deleting local files, checking for existing one first) 

Thanks for contributing!


isaacna:
I think the test is failing on windows because of NamedTemporaryFile not working on Windows: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

Lmk if there's a better way to test `remove_local_file`, but I can look into this later

cdp_backend/file_store/functions.py
```diff
@@ -0,0 +1,76 @@
...
+)
+log = logging.getLogger(__name__)
+
+GCS_URI = "gs://{bucket}/{filename}"
+
+###############################################################################
+
+
+def initialize_gcs_file_system(credentials_file: str) -> GCSFileSystem:
+    return GCSFileSystem(token=credentials_file)
```
JacksonMaxfield, line 25: Unfortunately we may need to have the credentials file as a param for get file uri and upload file and such just like I changed for the db upload functions.

Will see.

cdp_backend/file_store/functions.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def get_file_uri(fs: GCSFileSystem, bucket: str, filename: str) -> str:
```
JacksonMaxfield, line 28: Return type should be `Union[str, None]` I think?

cdp_backend/tests/file_store/test_functions.py
```diff
@@ -0,0 +1,107 @@
...
+)
+def test_remove_local_file() -> None:
+    f = NamedTemporaryFile()
+    f.write(b"Hello world!")
+
+    assert os.path.isfile(f.name)
+
+    functions.remove_local_file(f.name)
+
+    assert not os.path.isfile(f.name)
```
JacksonMaxfield, line 107: You should be able to replace this with the pytest `tmpdir` fixture: https://docs.pytest.org/en/stable/tmpdir.html#the-tmpdir-fixture

JacksonMaxfield (approved):
Looks good to me minus some minor comments.

Can we get docstrings for everything though?

cdp_backend/file_store/functions.py
isaacna: Good catch, I just changed it to Optional[str]

cdp_backend/file_store/functions.py
isaacna: Ah I see, I think we'll definitely have to do that if we do multiprocessing so I went ahead and made the change

PR #49, opened 2021-02-18
feature/event-gather-audio-creation

isaacna:
 - Part of #16 
 - Audio creation logic: essentially a prefect version of the logic from cdptools 
 - Removed the use of the `resource_exists` validator on the `File` db model because it doesn't work for `gs://` urls
 - Ran a manual test in addition to unit tests
 - changed `video_uri` in the example ingestion model because `external_resource_copy` was having trouble with cloning from the youtube url

Thanks for contributing!


cdp_backend/database/models.py
```diff
@@ -24,7 +24,7 @@ class File(Model):
     A file from the CDP file store.
     """
 
-    uri = fields.TextField(required=True, validator=validators.resource_exists)
+    uri = fields.TextField(required=True)
```
JacksonMaxfield, line 27: Hmmmm I think for now this is okay. There is a part of me though that wants to keep this. I think the way to do it would be to pass / accept `**kwargs` on the `resource_exists` function and if `fs` was provided as a kwarg use it instead of creating one?

This is assuming the issue you were encountering was that the created `fs` didn't have credentials to read the file existing or not.

cdp_backend/file_store/functions.py
```diff
@@ -142,3 +143,34 @@ def remove_local_file(filepath: Union[str, Path]) -> None:
...
+        credentials_file=credentials_file,
+        bucket=bucket,
+        filepath=filepath,
+        save_name=save_name,
+        remove_local=remove_local,
+    )
+
+
+@task
+def remove_local_file_task(filepath: Union[str, Path], trigger: Any) -> None:
```
JacksonMaxfield, line 166: `trigger`? I thought `trigger` parameters lived on the `@task` decorator? Or is this `mypy` complaining?

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -72,6 +77,14 @@ def create_event_gather_flow(
...
                 )
 
+                # TODO create/get transcript
+
+                # Create unique key for video uri
+                key = hashlib.sha256(session.video_uri.encode("utf8")).hexdigest()
+
+                # create/get audio (happens as part of transcript process)
+                create_or_get_audio(key, session.video_uri, bucket, credentials_file)
+
```
JacksonMaxfield, line 87: I understand the need to have `event_ref` and `body_ref` defined prior to the session loop but I think move the transcript processing to before the session database creation. Feels safer to ensure that all the heavy processing is completed without errors prior to uploading to database.

My dream change would be do to all the processing prior to uploading any database models. Which can be done by storing the results of processing in a temporary container then doing all the database calls with the results of the processing. That change can happen later though so your call on doing it in this PR or in a later one. That is a small enough (but nice amount of work) change that it may be nice to keep as an issue for new back-end devs to the project?

JacksonMaxfield: There is also a part of me that wants to hash the entire video file content as URIs are mutable whereas video content is at least a bit more static.

JacksonMaxfield: The specific case for hashing the file content instead of hashing the URL is that I can imagine a CDP instance that every day hits a page for like "latest from council" which only has one video but that video is updated every day. In such a case, the URL wouldn't change but the content would.

JacksonMaxfield: Again this is a small enough change that we can split out into it's own issue and have someone else take on if you like that too.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -139,3 +152,121 @@ def create_session_from_ingestion_model(
...
+    return Path(filepath).resolve(strict=True).name
+
+
+def create_or_get_audio(
+    key: str, video_uri: str, bucket: str, credentials_file: str
+) -> str:
+    tmp_audio_filepath = f"{key}_audio.wav"
+    audio_uri = fs_functions.get_file_uri_task(
+        bucket=bucket, filename=tmp_audio_filepath, credentials_file=credentials_file
+    )
```
JacksonMaxfield, line 177: Move these lines of code to below the docstring. Honestly surprised that black didn't complain about this haha.

cdp_backend/tests/pipeline/test_event_gather_pipeline.py
```diff
@@ -39,10 +46,56 @@ def test_create_event_gather_flow() -> None:
...
 
 
+@mock.patch("cdp_backend.file_store.functions.remove_local_file_task")
+@mock.patch("cdp_backend.database.functions.upload_db_model_task")
+@mock.patch("cdp_backend.pipeline.event_gather_pipeline.create_file")
+@mock.patch("cdp_backend.pipeline.event_gather_pipeline.create_filename_from_filepath")
+@mock.patch("cdp_backend.utils.file_utils.split_audio_task")
+@mock.patch("cdp_backend.utils.file_utils.external_resource_copy_task")
+@mock.patch("cdp_backend.file_store.functions.upload_file_task")
+@mock.patch("cdp_backend.file_store.functions.get_file_uri_task")
```
JacksonMaxfield, line 61: I haven't seen this functionality before, is this just a "hey pytest / unittest, just never run these functions?"

cdp_backend/tests/pipeline/test_event_gather_pipeline.py
```diff
@@ -39,10 +46,56 @@ def test_create_event_gather_flow() -> None:
...
+    mock_upload_db: MagicMock,
+    mock_remove_file: MagicMock,
+    key: str,
+    video_uri: str,
+    get_file_uri_value: str,
+    upload_file_value: str,
+    expected_audio_uri: str,
+) -> None:
+    mock_get_file_uri.return_value = get_file_uri_value
+    mock_external_copy.return_value = "stupid shit"
```
JacksonMaxfield, line 85: LOL. Uhhhhh lets change the value but I laughed.

JacksonMaxfield (commented):
Looks good to me, just some minor changes and comments for discussion before approve.

cdp_backend/database/models.py
isaacna: Sounds good I'll try that, and yeah the issue was credentials for `gs://` urls

cdp_backend/file_store/functions.py
isaacna: Oh I wrote this kind of confusingly, but `trigger` was just the return value of the upstream task that `remove_local_file_task` is dependent on, because otherwise prefect won't build the DAG properly and could remove the file when it's still needed. I can change this to be a more intuitive name like `upstream_done` or something similar.

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Sounds good, I'll move the audio/transcript creation to before the session loop. 

And I'll create separate issues for moving all the upload logic to after processing and hashing file content as opposed to the URL

cdp_backend/tests/pipeline/test_event_gather_pipeline.py
isaacna: Yeah, it's kinda difficult writing tests for the pipeline without a bunch of mocks, but this was much cleaner than multiple layers of "with mock.patch as x"

cdp_backend/tests/pipeline/test_event_gather_pipeline.py
isaacna: haha my bad, this is one of my habits when writing code but that's what reviews are for

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Created issues #50 and #51

cdp_backend/file_store/functions.py
JacksonMaxfield: Ahhh I see. You don't actually need this. On all tasks there is a `upstream_tasks` kwarg. https://docs.prefect.io/api/latest/triggers.html#triggers

So as long as you just specify:

```python
some_upstream_result = some_upstream_processing()
remove_local_file("file.png", upstream_tasks=[some_upstream_result])
```

it will work just like this but without the need for documenting the parameter.

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: Uhhhhh reminder, the audio / transcript processing is on a per-session basis, so keep these loops for now, but yea _eventually_ as a part of the created issues is to do basically one processing loop, then one upload loops.

cdp_backend/file_store/functions.py
isaacna: Oh dope I was looking for something like that but couldn't find it in the docs

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Yeah my bad, meant to say I'll do it before the uploads, but still in the session loop

cdp_backend/database/models.py
isaacna: Actually after trying this, I think passing another param to the validator is a little tricky. 

I can add kwargs to `resource_exists`, but I'm not sure if there's an easy way to pass in `fs` into that function inside `uri = fields.TextField(required=True, validator=validators.resource_exists)`. 

Because model fields in FireO are assigned like `db_file.uri = "gs://stuff"`, I'm not sure we can pass in other params into the `validators` arg. 

Would it make more sense to just validate separately outside of FireO's built-in field validation?

JacksonMaxfield: Yea.... Hm. I guess get rid of it for now. I will add an issue to the repo for tracking it as a "potentially add back" whenever we think of a solution.

JacksonMaxfield: Added as issue: https://github.com/CouncilDataProject/cdp-backend/issues/52

Feel free to proceed without this validator. Marking this conversation as resolved.

JacksonMaxfield (approved):
One minor comment otherwise everything looks good to me! Thanks for making all the changes.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -139,3 +152,126 @@ def create_session_from_ingestion_model(
...
+    db_file = db_models.File()
+    db_file.name = name
+    db_file.uri = uri
+
+    return db_file
+
+
+@task
+def create_filename_from_filepath(filepath: str) -> str:
+    return Path(filepath).resolve(strict=True).name
```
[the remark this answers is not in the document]
isaacna: Sounds good, I'll move `create_filename_from_filepath` to `filestore/functions.py`, and all the other functions that create db models from some input into `database/functions.py`. 

I originally put them here since I they weren't necessarily functions that explicitly interacted with the filestore or database, but the event gather pipeline file is getting kinda big but I agree this will also help move out some of the logic

PR #54, opened 2021-04-15
feature/add-sr-models

isaacna:
- [ ] Link to any relevant issue in the PR description. Ex: _Resolves [gh-12], adds tiff file format support_
 - Adding SR models from cdptools which is required for transcript creation in event gather pipeline
 - Fixed mypy issues (a lot of the typing mypy issues are pretty arbitrary IMO where I added `type: ignore`
 - Had to modify the model args a bit as mypy requires the ABC and implementations to have the exact same arguments

Thanks for contributing!


cdp_backend/sr_models/constants.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from typing import Dict, List, Union
+
+# Transcript data json blobs will look like the following
+# [{"start_time": 0.0, "portion": "Hello world!", "end_time": 2.1}, ...]
+TranscriptDataJSON = List[Dict[str, Union[str, float]]]
```
JacksonMaxfield, line 8: There is a `TextBlockData` object in the repo we should probably use: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/transcript_model.py#L16

cdp_backend/sr_models/constants.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class TranscriptFormats:
+    raw = "raw"
+    timestamped_words = "timestamped-words"
+    timestamped_sentences = "timestamped-sentences"
+    timestamped_speaker_turns = "timestamped-speaker-turns"
```
JacksonMaxfield, line 15: I am somewhat tempted to just cut this and say we only support timestamped words. Which would result in `TextBlockData` changes probably (it's fine as is just property renaming). This is from thinking about how LocalVoicesNetwork manage their transcript data.

cdp_backend/sr_models/sr_model.py
```diff
@@ -0,0 +1,104 @@
...
+        storage_ready: Dict[str, Union[str, float, List[Dict[str, Any]],
+        constants.TranscriptDataJSON]]
+            The in memory version of a transcript file about to be dumped to JSON.
+        """
+        return {
+            "format": transcript_format,
+            "annotations": annotations,
+            "confidence": confidence,
+            "data": data,
+        }
```
JacksonMaxfield, line 58: This is really just handled by the `TextBlockData` and `Transcript` models now... (i think)

cdp_backend/sr_models/sr_model.py
```diff
@@ -0,0 +1,104 @@
...
+
+###############################################################################
+
+
+class SRModelOutputs(NamedTuple):
+    raw_path: Path
+    confidence: float
+    timestamped_words_path: Optional[Path] = None
+    timestamped_sentences_path: Optional[Path] = None
+    timestamped_speaker_turns_path: Optional[Path] = None
```
JacksonMaxfield, line 18: Again... thinking about removing extras for just only in favor of words.

JacksonMaxfield: I really need to look into the LVN transcript spec a bit more.

JacksonMaxfield (commented):
Alright so I have a lot of thoughts on this + transcript stuff in general and I think I forgot how much craziness is in the SR model code.

I am going to do some thinking on the transcript format a bit (comparing against the LVN transcript spec).

I recall you said you were getting really busy so I would say "don't worry about this PR" I will ping you if I have questions but I think some thinking needs to be put in on my side unfortunately. And I will post updates as I go if you want to review.

Thoughts?

isaacna:
> Alright so I have a lot of thoughts on this + transcript stuff in general and I think I forgot how much craziness is in the SR model code.
> 
> I am going to do some thinking on the transcript format a bit (comparing against the LVN transcript spec).
> 
> I recall you said you were getting really busy so I would say "don't worry about this PR" I will ping you if I have questions but I think some thinking needs to be put in on my side unfortunately. And I will post updates as I go if you want to review.
> 
> Thoughts?

Sounds good, I could get started on the transcript creation part of the pipeline without actually using sr model code until this is done. And sure lmk if you have any questions but I didn't do too much thinking about the actual logic for this, I mostly just brought it over from cdptools and fixed the checkstyle errors

JacksonMaxfield:
So I got some updates / ideas:

1. You can drop the `WebVTTSRModel` for this PR, as it will be a lot of changes to get it working with my following recommendations.

2. I think we should only allow timestamped words, so no need to have multiple transcript versions so can get rid of that as well (should simplify a lot of the pipeline + code too!).

3. I like how LVN structures _some_ of their data. Not all, but some. Currently our v3 transcript model is defined [here](https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/transcript_model.py#L46) but I think we should slightly change it. I can make this PR or you can include in this PR. My suggestion (and would love feedback):
```json
{
    "confidence": float,
    "generator": str,
    "session_datetime": str,
    "created_datetime": str,
    "sentences": List[Sentence],
    "annotations": Dict[str, Any],
}
```

And we define a new `Sentence`:
```json
{
    "index": int,
    "confidence": float,
    "start_time": float,
    "end_time": float,
    "speaker": Optional[str],
    "annotations": Dict[str, Any],
    "words": List[Word],
}
```

And we finally define `Word`:
```json
{
    "index": int,
    "start_time": float,
    "end_time": float,
    "text": str,
    "annotations": Dict[str, Any],
}
```

Any place where we have an `index` key and value I would try to keep them sorted but it's just extra metadata so I say "why not keep it." This model basically locks us into sentence and word models which I think is totally fine. It brings us _slightly_ closer to the LVN model (but not really), the only change from our base model is just the explicit definition of sentences and words instead of just `TextBlockData`.

4. With all that said. I think the todos / general thoughts for this are:
* [ ] Drop `WebVTTSRModel` (for now)
* [ ] Because we have a model we can probably get rid of the function: `wrap_and_format_transcript_data` in the `SRModel` base class.
* [ ] The return value from the `transcribe` function should just be a single path? Or the `Transcript` model object, I think I like returning the whole `Transcript` model because then you can get rid of the `save_path` param entirely and because I think we store certain values from the transcript in the database aftewards so we wouldn't need to load it _back into memory_.
* [ ] Make the above changes basically just for the `GoogleSRModel`.

Thoughts? Ideas? Complaints?

isaacna:
This works for me, I think moving over to this data format makes sense. And I like the idea of returning the whole `Transcript` object, since the save path paradigm felt a little awkward to me. 

I'll get started on these changes!

isaacna:
I noticed that in the current google sr model code, each sentence's content is just a string made up of each word, however in the new data model we're working with, `Sentence` contains a list of `Word`. 

Should we also add a `joined_words` field or something similar in `Sentence` for easier use, rather than having whatever consumes this class have to process this nested model? Not sure how we actually plan to use the model, but just an option if we want to make the sentence model slightly easier to use (in exchange for some data redundancy)

JacksonMaxfield:
Yea I get that and thought about it as well but the thing that swayed me was that the front-end loads these files on client side so reducing / minimizing the size of the transcript is always in our interest. The one thing that I hadn't thought about was how to handle sentence punctuation, i.e. commas splitting sentences, periods, quotes, etc. Are those characters embedded in the word strings?

i.e. is it:
```
{
    "index": 0,
    "start_time": 101.2,
    "end_time": 101.5,
    "text": "blah,",
    "annotations": {},
}
```

or 
```
{
    "index": 0,
    "start_time": 101.2,
    "end_time": 101.5,
    "text": "blah",
    "annotations": {},
}
```

(no comma on the end of "blah")?

What I think we could do there is to have the "words" be _only_ the "non-deliminating" characters. I.e. everything except `a-zA-Z0-9'\-` (apostrophe and literal hyphen included) gets stripped before storage in the word model. And then have a field on the Sentence model that is "text" that is the full text of the sentence including the deliminating characters?

isaacna:
I can't say for sure whether the punctuation is attached to the words or not since I've never actually run this code. But based off [this snippet](https://github.com/CouncilDataProject/cdp-backend/pull/54/files#diff-2bd83fe4f0d6b266c79ac880dbeda1f75bac1d4e535b2e8ffffc1aabc2560f14R141-R145) in the existing code I'm assuming so?

And that idea of stripping the characters works for me

JacksonMaxfield:
Yeah there is punctuation attached to the return from GCP. I will dump the raw return of speech to text to JSON and send over this weekend.

isaacna:
Here's an example of the [transcript](https://pastebin.com/ep4NipFL) generated for this [council meeting](https://video.seattle.gov/media/council/brief_072219_2011957V.mp4)

cdp_backend/pipeline/transcript_model.py
```diff
@@ -13,32 +13,68 @@
...
     ----------
+    index: int
+        The index of the word in it's respective sentence.
+    start_time: float
+        Time in seconds for when this word begins.
+    end_time: float
+        Time in seconds for when this word ends.
     text: str
-        The full text portion of this data block.
+        The text of the word, cleaned of all non-deliminating chars.
```
JacksonMaxfield, line 29: Can we make this lowercased as well? Seems like a giant list of the words will basically then just act as the "raw" text.

cdp_backend/pipeline/transcript_model.py
```diff
@@ -13,32 +13,68 @@
...
     annotations: Optional[Dict[str, Any]]
-        Any annotations specific to this text block.
+        Any annotations specific to this sentence.
         Default: None (no annotations)
-        See TextBlockAnnotations for list of known annotations and their dtypes.
+    words: List[Word]
+        The list of word for the sentence.
+        See Word model for more info.
+    text: str
+        The text of the sentence including non-deliminating chars.
```
JacksonMaxfield, line 67: I might expand this to say: "The text of the sentence including all formatting and non-deliminating chars." To handle any "lower case vs upper case" style stuff too.

cdp_backend/pipeline/transcript_model.py
```diff
@@ -13,32 +13,68 @@
...
 
-    text: str
+    index: int
+    confidence: float
     start_time: float
     end_time: float
-    speaker: Optional[str] = None
-    annotations: Optional[Dict[str, Any]] = None
+    speaker: Optional[str]
+    annotations: Optional[Dict[str, Any]]
```
JacksonMaxfield, line 75: It's probably a Python thing I am forgetting... but why were these defaults removed?

cdp_backend/pipeline/transcript_model.py
```diff
@@ -58,13 +94,13 @@ class Transcript:
...
         Example: "Google Speech-to-Text -- Lib Version: 2.0.1"
-    session_datetime: str
+    session_datetime: Optional[str]
         ISO formatted datetime for the session that this document transcribes.
     created_datetime: str
         ISO formatted datetime for when this transcript was created.
-    data: List[TextBlockData]
-        A list of text blocks.
-        See TextBlockData documentation for more information.
+    sentences: List[Sentences]
```
JacksonMaxfield, line 101: Very minor nit: typing in doc should be `List[Sentence]` -- extra trailing `s`

cdp_backend/pipeline/transcript_model.py
```diff
@@ -87,9 +123,9 @@ class Transcript:
 
     confidence: float
     generator: str
-    session_datetime: str
+    session_datetime: Optional[str]
     created_datetime: str
-    data: List[TextBlockData]
+    sentences: List[Sentence]
     annotations: Optional[Dict[str, Any]] = None
```
JacksonMaxfield, line 129: Comment for myself... I am tempted to include more metadata here. But I think we can do that in annotations or something. :thinking: 

Things like "source" etc. I feel like source should be included.... The archivists would want that.

cdp_backend/sr_models/google_cloud_sr_model.py
```diff
@@ -0,0 +1,181 @@
...
+            return cleaned
+        else:
+            return []
+
+    def transcribe(
+        self,
+        file_uri: Union[str, Path],
+        phrases: Optional[List[str]] = None,
+        **kwargs: Any,
+    ) -> Transcript:
```
JacksonMaxfield, line 54: Can you add a docstring here.

Iirc, file_uri should be a GCS file uri (`gs://...`) and phrases is a list of str things that make the model perform better. I.e. if we drop in the minutes items the model will have a better chance of picking up the various "political language" used.

cdp_backend/sr_models/google_cloud_sr_model.py
```diff
@@ -0,0 +1,181 @@
...
+                        cleaned_phrase = cleaned_phrase[: cleaned_phrase.rfind(" ")]
+
+                    # Append cleaned phrase and increase character count
+                    cleaned.append(cleaned_phrase)
+                    total_character_count += len(cleaned_phrase)
+
+            return cleaned
+        else:
+            return []
+
```
JacksonMaxfield, line 48: I know this was my code originally but nitpick:

this can just be `return []` no need to wrap in an else block.

cdp_backend/sr_models/google_cloud_sr_model.py
```diff
@@ -0,0 +1,181 @@
...
+        log.debug(f"Beginning transcription for: {file_uri}")
+        operation = client.long_running_recognize(config, audio)
+
+        # Wait for complete
+        response = operation.result(timeout=10800)
+
+        # Select highest confidence transcripts
+        confidence_sum = 0
+        segments = 0
+        # timestamped_words = []
```
JacksonMaxfield, line 91: Remove commented out

cdp_backend/sr_models/google_cloud_sr_model.py
```diff
@@ -0,0 +1,181 @@
...
+                        end_time = word.end_time.seconds + word.end_time.nanos * 1e-9
+
+                        # Clean everything but non-delimiting characters
+                        regex = re.compile(r"[^a-zA-Z0-9'\-]")
+                        cleaned_word = regex.sub("", word.word)
+                        timestamped_word = Word(
+                            index=word_index,
+                            start_time=start_time,
+                            end_time=end_time,
+                            text=cleaned_word,
```
JacksonMaxfield, line 119: Can you `cleaned_word.lower()` here as discussed above.

cdp_backend/sr_models/google_cloud_sr_model.py
```diff
@@ -0,0 +1,181 @@
...
+                                speaker=None,
+                                annotations=None,
+                                words=[timestamped_word],
+                                text=word.word,
+                            )
+                            word_index += 1
+
+                        # End current sentence and reset
+                        # TODO: Account for non-sentence ending periods, such as
+                        # prefixes like "Mr." or "Dr."
```
JacksonMaxfield, line 140: Text data is HARD.

cdp_backend/sr_models/sr_model.py
JacksonMaxfield: Bump. Can remove this :slightly_smiling_face:

cdp_backend/sr_models/sr_model.py
```diff
@@ -0,0 +1,44 @@
...
+    extras: Optional[Dict[str, Any]] = None
+
+
+class SRModel(ABC):
+    @abstractmethod
+    def transcribe(
+        self,
+        file_uri: Union[str, Path],
+        phrases: Optional[List[str]] = None,
```
JacksonMaxfield, line 27: If all the typing is correct `phrases` shouldn't need to be on the ABC function def.

JacksonMaxfield (commented):
This looks so good. Thanks for making all the changes. Some minor nitpicks and comments in general.

JacksonMaxfield:
Also thanks for running a meeting through the system. Really helped with seeing the produced object :+1:

cdp_backend/pipeline/transcript_model.py
isaacna: I actually can't remember, I think I may have done this for some mypy error but I just moved these defaults as the last params of the object and that works

cdp_backend/sr_models/google_cloud_sr_model.py
isaacna: Yeah, there are a few hacky ways I think we could address this but imo there might just be some other library out there that can better detect edge cases for sentence ending. Not super necessary right no imo but food for thought

cdp_backend/sr_models/sr_model.py
isaacna: I'm a little confused by this, if `phrases` is on the function signature in `GoogleCloudSRModel`, wouldn't we need this in the abstract method? Mypy complains if the implementation of an ABC function doesn't have the exact same function signature, even if it's a superset of the ABC function

isaacna: If we don't want it in the ABC function, I think we could remove it and pass it in as part of `kwargs` if you think that's more generalized for a `transcribe` method

cdp_backend/sr_models/google_cloud_sr_model.py
JacksonMaxfield: Spacy has us covered: https://stackoverflow.com/questions/46290313/how-to-break-up-document-by-sentences-with-with-spacy

cdp_backend/sr_models/sr_model.py
JacksonMaxfield: Interesting..... it should be covered by kwargs though....

JacksonMaxfield: I have done this exact thing in the imaging library I maintain:

ABC class init with single param and kwargs: https://github.com/AllenCellModeling/aicsimageio/blob/main/aicsimageio/readers/reader.py#L111

inheritting class with extra params: https://github.com/AllenCellModeling/aicsimageio/blob/main/aicsimageio/readers/tiff_reader.py#L67

JacksonMaxfield: I wonder if it's the typing on the ABC kwargs?

isaacna: Lol I just tried again and mypy seems to be fine with it so I'll remove `phrases` from the SR model. Not sure what was going on before but maybe there was some other issue with the model

cdp_backend/sr_models/google_cloud_sr_model.py
isaacna: Oh this is pretty cool, I'll create an issue for integrating spacy into the transcription

isaacna: https://github.com/CouncilDataProject/cdp-backend/issues/56

cdp_backend/pipeline/transcript_model.py
```diff
@@ -137,40 +173,59 @@ class TextBlockAnnotations(Enum):
...
-                TextBlockAnnotations.confidence.name: 0.999,
-            },
-        ),
-        TextBlockData(
-            text="Who is speaking this?",
-            start_time=2.2,
-            end_time=3.9,
-            annotations={
-                TextBlockAnnotations.confidence.name: 0.835,
+                TextBlockAnnotations.confidence.name: 0.987,
```
JacksonMaxfield, line 228: Last nitpick and then good to go, the text in the `Word` objects in this example should now be lower case.

JacksonMaxfield (approved):
Single last nitpick then happy to merge :slightly_smiling_face:

cdp_backend/pipeline/transcript_model.py
isaacna: Good catch, fixed

Issue #14, opened 2020-11-09
Set public read during CDPStack creation


evamaxfield:
## Use Case
*Please provide a use case to help us understand your request in context*

Maybe this should be a parameter in the CDPStack object but currently our standard for infrastructure is to allow public read so during stack creation, we should set the default read permissions for the bucket and database to be public as well.

I say potentially a parameter because dev stacks don't need to be public read which could save money for the dev :upside_down_face: 

(the dev stack with example data should really only cost like $0.01 though since the example data should be really minimal)

## Solution
*Please describe your ideal solution*

I haven't figured out which Pulumi resource is needed (or really if it is possible to do with the pulumi API, it may need to be a second `ComponentResource` that calls out to `gcloud` in a subprocess)

But setting this and managing permissions through code would be incredibly useful.

## Alternatives
*Please describe any alternatives you've considered, even if you've dismissed them*

JacksonMaxfield:
Seems like we can use the [Firestore CLI](https://cloud.google.com/firestore/docs/security/get-started#use_the_firebase_cli) to do this. So either a post-script thing _or_ writing our own `ComponentResource` to do subprocess it out to the CLI

JacksonMaxfield:
This is now possible with the Pulumi 3.0 Google Native Provider:

https://www.pulumi.com/docs/reference/pkg/google-native/firebaserules/v1/ruleset/

I attempted it but it is finicky and tricky to get right / I may have been doing something incorrectly. Going to hold off until needed for research queries.

JacksonMaxfield:
Firestore public read rules:
```
rules_version = '2';
service cloud.firestore {
  match /databases/{database}/documents {
    match /{document=**} {
      allow read: if request.auth == null;
    }
  }
}
```

Firebase storage public read rules:
```
rules_version = '2';
service firebase.storage {
  match /b/{bucket}/o {
    match /{allPaths=**} {
      allow read;
    }
  }
}
```

JacksonMaxfield:
Looks like the process for this will be:

1. create the rulesets with: https://www.pulumi.com/docs/reference/pkg/google-native/firebaserules/v1/ruleset/
2. use the rulesets in https://www.pulumi.com/docs/reference/pkg/google-native/firebaserules/v1/release/

JacksonMaxfield:
Got it working! :tada:

I ditched trying to get Pulumi ruleset working and went with a simple node script

JacksonMaxfield:
[...]

PR #55, opened 2021-04-29
feature/pulumi-3.0

JacksonMaxfield:

W.I.P for https://github.com/CouncilDataProject/cdp-backend/issues/14#issuecomment-828917442

One of the issues with the current pulumi stack configuration was how Cloud Firestore indexes were being managed. Because we didn't have full access over the resource configuration, every time pulumi ran it would think that the indexes were different: [See CouncilDataProject/example Infra Deploy CI/CD](https://github.com/CouncilDataProject/example/actions?query=workflow%3A%22Infrastructure%22).

With Pulumi 3.0 they added the "Google Cloud Native" provider which is basically a direct translation from the Google Cloud API to their resources instead of manually mantained and created code. Both are useful and both are used in this PR but for things that require deep control, such as database index creation, the native provider is a bit better as it will know that nothing has changed between deployments for indexes (Yayyyy no more failing infra builds).

Thanks for contributing!

### DIFF
--- a/cdp_backend/bin/run_cdp_event_gather.py
+++ b/cdp_backend/bin/run_cdp_event_gather.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 import argparse
+import json
 import logging
 import sys
 import traceback
@@ -54,7 +55,10 @@ class Args(argparse.Namespace):
             "--google_cloud_storage_bucket_name",
             type=Path,
             dest="gcs_bucket",
-            help=("Google Cloud Storage bucket name for file uploads."),
+            help=(
+                "Google Cloud Storage bucket name for file uploads. "
+                "Default: None (assume bucket name from GCP project id)"
+            ),
         )
 
         p.parse_args(namespace=self)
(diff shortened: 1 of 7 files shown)


cdp_backend/bin/run_cdp_event_gather.py
```diff
@@ -71,11 +75,18 @@ def main() -> None:
...
         credentials_file = args.google_credentials_file
-
         get_events_func = import_get_events_func(args.get_events_function_path)
 
-        bucket = args.gcs_bucket
+        # Handle default None bucket
+        if args.gcs_bucket is None:
+            with open(credentials_file, "r") as open_resource:
+                project_id = json.load(open_resource)["project_id"]
+                bucket = f"{project_id}.appspot.com"
```
isaacna, line 86: Just curious but all GCP buckets use "appspot.com" by default right? I know mine does but just wanted to double check

dev-infrastructure/Makefile
```diff
@@ -21,29 +21,28 @@ login: ## Run GCloud and Pulumi logins
...
-	make login
 	gcloud projects create $(project) --set-as-default
 	pulumi stack init $(project)
 	echo "----------------------------------------------------------------------------"
 	echo "Follow the link to setup billing for the created GCloud account."
 	echo "https://console.cloud.google.com/billing/linkedaccount?project=${project}"
 
 build: ## Run pulumi up for infra setup
-	pulumi up -p 5
+	pulumi up -p 4
```
isaacna, line 31: For my own knowledge, what's the difference between `4` and `5` as the arg here?

isaacna (approved):

cdp_backend/bin/run_cdp_event_gather.py
JacksonMaxfield: Nope, just ones connected to firebase / firestore. We could create a bucket entirely independent of firebase / firestore and it wouldnt have that suffix.

dev-infrastructure/Makefile
JacksonMaxfield: how many resources can be deployed in parallel. I was noticing some oddities when testing this with API limits so dropped it down even further.

PR #59, opened 2021-05-31
Feature/use spacy in transcription

### LINKED ISSUE #56: Use spacy to separate sentences in SR model transcribe function
## Use Case
To account for edge cases in sentence detection (Prefixes like Mr.), we can use the [spacy](https://stackoverflow.com/questions/46290313/how-to-break-up-document-by-sentences-with-with-spacy) library. 

## Solution
Use the spacy library to separate sentences. Haven't looked into the library too deeply, but would probably require us to change the iteration logic since we create `Word` (and increment index) and potentially end the `Sentence` in the same iteration.

isaacna:
 - Resolves #56 
 - Integrating spaCy library into SR model transcription to have better sentence boundary detection
 - I tested spaCy with some harder edge cases (having things like "Mr." in the sentence) and works great. 
 - There are a [few different types of sentence segmenters](https://spacy.io/usage/linguistic-features#sbd), but I think "Sentencizer" works just fine for our use case
 - Pastebin with printed `Transcript` document: https://pastebin.com/iP8SydpH

Thanks for contributing!

### DIFF
--- a/cdp_backend/sr_models/google_cloud_sr_model.py
+++ b/cdp_backend/sr_models/google_cloud_sr_model.py
@@ -206,6 +206,4 @@ class GoogleCloudSRModel(SRModel):
             annotations=None,
         )
 
-        print(transcript)
-
         return transcript
--- a/cdp_backend/sr_models/google_cloud_sr_model.py
+++ b/cdp_backend/sr_models/google_cloud_sr_model.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import Any, List, Optional, Union
 
 from google.cloud import speech_v1p1beta1 as speech
+from spacy.lang.en import English
 
 from ..pipeline.transcript_model import Sentence, Transcript, Word
 from .sr_model import SRModel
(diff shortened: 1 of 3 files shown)


cdp_backend/sr_models/google_cloud_sr_model.py
```diff
@@ -107,72 +108,86 @@ def transcribe(
...
+                    # Sentence text
+                    s_text = sentences[s_ind]
+
+                    num_words = len(s_text.split())
+
+                    # Initialize sentence model
+                    timestamped_sentence = Sentence(
+                        index=transcript_sentence_index,
+                        confidence=result.alternatives[0].confidence,
+                        # Start and end time are laceholder values
```
JacksonMaxfield, line 139: typo: `laceholder` -> `placeholder`

setup.py
```diff
@@ -58,8 +58,11 @@
     "pulumi~=3.0.0",
     "pulumi-google-native~=0.1.0",
     "pulumi-gcp~=5.0.0",
+    "setuptools~=57.0.0",
```
JacksonMaxfield, line 61: A bit confused as to why this is needed.... look at the first line of this file, _we are importing setuptools_. And `wheel` is stdlib I thought.

JacksonMaxfield: Can you push a commit without `setuptools` and `wheel` just to see the error.

JacksonMaxfield (approved):
This is awesome. Really great logic / parsing changes too!

Typo + debugging dep install stuff then good to go.

setup.py
isaacna: I added these because the documentation said to install them before installing spacy, but yeah those were redundant so I can remove them

PR #64, opened 2021-06-27
feature/add-webvtt-sr-model

JacksonMaxfield:

W.I.P. #16 

Brings in the WebVTTSRModel to the library. Some major changes from `cdptools`. Generally it's a bit simpler of a function in that there are overall less functions but a bit more looping because of how I decided to do the truecasing and sentence splitting.

I used the Sentencizer that @isaacna found and used.

I also changed the transcript model itself... @tohuynh you should be aware of these changes for the front-end.
I changed the `speaker` property to `speaker_name` and added a `speaker_index`. This is useful both for just finding which sentences belong to speaker blocks but also in the case that we have multiple speaker blocks with the same speaker name. Generally this would be rare except in the obvious case of "Unknown Speaker" or similar.

Generated the short transcript from the already stored webvtt file in the test resources. See the contents of that JSON file as it's really nice imo.

Thanks for contributing!


cdp_backend/sr_models/webvtt_sr_model.py
```diff
@@ -0,0 +1,218 @@
...
+                            text=self._clean_word(word),
+                            # Temp raw word storage
+                            # will remove after we truecase and sentenceizer
+                            annotations={"raw": word},
+                        )
+                    )
+
+            all_speaker_sentences.append(
+                Sentence(
+                    index=0,
```
tohuynh, line 148: I was a little confused, here, you are using `Sentence` to represent possibly a speaker block (it can contain 1 or more true English sentences), right?

tohuynh (commented):
Thanks for letting me know about the changes. Would `speaker_index` be used in the front-end later?

Looks good overall, just a couple of questions.

cdp_backend/sr_models/google_cloud_sr_model.py
```diff
@@ -125,8 +123,7 @@ def transcribe(
 
                 # Index holder for word results of response
                 w_marker = 0
-
-                for s_ind in range(0, len(sentences)):
+                for s_ind, _ in enumerate(sentences):
```
isaacna, line 126: I did not know Python had this but good to know

cdp_backend/sr_models/webvtt_sr_model.py
```diff
@@ -0,0 +1,218 @@
...
+
+        # Confidence is tricky. We allow it to be a parameter because closed captions
+        # aren't always 100% accurate. For Seattle, I would guess they are about 97%
+        # accurate.
+        self.confidence = confidence
+
+    def _normalize_text(self, text: str) -> str:
+        normalized_text = truecase.get_true_case(text)
+
+        # Fix some punctuation issue, e.g. `roughly$ 19` bececomes `roughly $19`
```
isaacna, line 46: Are there any punctuation issues that could cause issues with sentence segmentation? I didn't notice anything weird with the google SR model when I was testing with it but wasn't sure about webvtt

cdp_backend/sr_models/webvtt_sr_model.py
```diff
@@ -0,0 +1,218 @@
...
+                if new_speaker_search:
+                    # Add previous speaker captions to all speaker captions
+                    # (and reset current speaker captions)
+                    if current_speaker_captions is not None:
+                        all_speaker_captions.append(current_speaker_captions)
+                        current_speaker_captions = None
+
+                # Create new current speaker
+                if current_speaker_captions is None:
+                    current_speaker_captions = [caption]
```
isaacna, line 92: Is there ever a case where `current_speaker_captions` would be `None` when `new_speaker_search` isn't true? Could we just replace line 88 with `current_speaker_catpions = [caption]`

isaacna (approved):

cdp_backend/sr_models/webvtt_sr_model.py
JacksonMaxfield: I don't believe so. I didn't notice anything either but I think this is one of things that we will have to adjust as we go.

cdp_backend/sr_models/webvtt_sr_model.py
JacksonMaxfield: I think so.... Would just have to change the logic to if len() == 0.

This check was to make mypy happy but I like making it always a list and checking length

cdp_backend/sr_models/webvtt_sr_model.py
```diff
@@ -0,0 +1,218 @@
...
+                                # Linear words per second
+                                # offset by start time of caption
+                                caption.start_in_seconds
+                                + ((i / len(raw_words)) * caption_duration)
+                            ),
+                            end_time=(
+                                # Linear words per second
+                                # offset by start time of caption + 1
+                                caption.start_in_seconds
+                                + (((i + 1) / len(raw_words)) * caption_duration)
```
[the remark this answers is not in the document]
JacksonMaxfield: Yes on the word level but never on the sentence start level I believe. And to be fair I already previously noticed that closed caption meetings had weird offsets sometimes.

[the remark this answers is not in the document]
JacksonMaxfield: This should be an okay method though. Linear interpolation in this case is fine I think

cdp_backend/sr_models/webvtt_sr_model.py
JacksonMaxfield: Yes. Just once to do a second rough of processing later. I can add more docs about this if you want as I know it's a bit confusing but it just made sense to me as to how to store all the required processing info. I may make it a custom namedtuple though to avoid confusion.

JacksonMaxfield:
> Thanks for letting me know about the changes. Would `speaker_index` be used in the front-end later?

Yes. On the front end in order to group sentences by speaker (unless I'm misremembering the UI) would be to group by index and then display each group.

cdp_backend/sr_models/webvtt_sr_model.py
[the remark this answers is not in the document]
tohuynh: On a sentence level too, I think if the sentence doesn't end at end of a caption (not sure if this ever happens for Seattle though). But from what I've seen the captions are kind of short in terms of the number of words per caption.

cdp_backend/sr_models/webvtt_sr_model.py
tohuynh: I think just a short comment will do.

tohuynh:
> Yes. On the front end in order to group sentences by speaker (unless I'm misremembering the UI) would be to group by index and then display each group.

I thought it was one sentence per transcript item from UI wireframes, but I do like one speaker block per transcript item more.

Issue #65, opened 2021-06-29
Generate thumbnail PNG and GIF for each video


JacksonMaxfield:
## Use Case
*Please provide a use case to help us understand your request in context*

Using our card example on the frontend documentation: https://councildataproject.org/cdp-frontend/?path=/story/library-cards-events--meeting-search-result

We need a thumbnail for each event. And we planned for this in two ways, our event schema in our database has both a static and hover thumbnail. https://councildataproject.org/cdp-backend/cdp_backend.database.html#cdp_backend.database.models.Event

## Solution
*Please describe your ideal solution*

* [ ] Write a function for the static thumbnail generation. Takes in a video file path, grabs a frame from the 30 second mark, saves to PNG, returns the save path.
* [ ] Write a function for the hover thumbnail generation. Takes in a video file path, grabs 10 frames from every 1/10th of the video file. So if the video is 1 hour long, it would grab a frame from every 6 minutes. Saves those frames to a GIF and returns the save path.

[imageio](https://imageio.readthedocs.io/en/stable/) is really all that is needed here.

## Alternatives
*Please describe any alternatives you've considered, even if you've dismissed them*

This can become much more complicated later. But for right now we really just need to get work done. The 30 second and the every 1/10th of the video file are chosen some what randomly but they should be okay. Down the line we may want have some better frame selection in place but not now.

After these two functions are written, we will need to add them to the pipeline. I am envisioning that a single prefect task would do the following:

```python
def generate_thumbnails(video_uri: str) -> Tuple[str, str]:
    # Download video
    local_video_copy_path = external_resource_copy(video_uri)

   # Process to get thumbnails
    static_thumbnail_path = get_static_thumbnail(local_video_copy_path)
    hover_thumbnail_path = get_hover_thumbnail(local_video_copy_path)

   # Store files in storage
    static_thumbnail_uri = upload_file(static_thumbnail_path)
    hover_thumbnail_uri = upload_file(hover_thumbnail_path)
    
    # Store file URIs in DB
    static_thumbnail_ref = upload_file_db(static_thumbnail_uri)
    hover_thumbnail_ref = upload_file_db(hover_thumbnail_uri)

    return static_thumbnail_ref, hover_thumbnail_ref
```

PR #68, opened 2021-07-08
feature/thumbnail-generation-functions

Shak2000:
- [X] Thumbnail Generator
- [X] Relevant Issue: [#65](https://github.com/CouncilDataProject/cdp-backend/issues/65)
- [X] The system needed a way of generating static and hover thumbnails.
- [X] The tests can be found [here](https://github.com/Shak2000/cdp-backend/blob/feature/thumbnail-generator/cdp_backend/tests/pipeline/test_thumbnail_generator.py).
- [X] No external documentation is currently needed.

Thanks for contributing!


cdp_backend/tests/pipeline/test_thumbnail_generator.py
```diff
@@ -0,0 +1,97 @@
...
+from pathlib import Path
+
+import pytest
+
+from cdp_backend.pipeline.thumbnail_generator import (
+    get_hover_thumbnail,
+    get_static_thumbnail,
+)
+
+test_data_directory = "./cdp_backend/tests/data/"
```
JacksonMaxfield, line 11: You can remove this! Instead of using this variable in all the test functions you can just add a parameter called `resources_dir` see an example here: https://github.com/CouncilDataProject/cdp-backend/blob/4fb864281ec8d822b303a40b25aaf6a133ed88d1/cdp_backend/tests/utils/test_file_utils.py#L20

That is just a "fixture" but the same applies to all tests. You can just use that parameter `resources_dir` and get the test files :slightly_smiling_face:

cdp_backend/tests/pipeline/test_thumbnail_generator.py
```diff
@@ -0,0 +1,97 @@
...
+Static thumbnail generator tests
+
+The static thumbnail generator produces a png thumbnail for an mp4 video
+"""
+
+
+# Test if the static thumbnail generator can produce a thumbnail with the default
+# timing
+# The default timing is always 30 seconds into the video
+def test_get_static_thumbnail_standard() -> None:
```
JacksonMaxfield, line 23: My biggest comment here is that all of these individual tests should be converted to a single test that is `@pytest.mark.parametrize` -- best example here: https://github.com/CouncilDataProject/cdp-backend/blob/4fb864281ec8d822b303a40b25aaf6a133ed88d1/cdp_backend/tests/utils/test_file_utils.py#L50

cdp_backend/tests/pipeline/test_thumbnail_generator.py
```diff
@@ -0,0 +1,97 @@
...
+def test_get_static_thumbnail_not_mp4() -> None:
+    with pytest.raises(Exception):
+        get_static_thumbnail(test_data_directory + "fake_creds.json")
+
+
+"""
+Hover thumbnail generator tests
+
+The static thumbnail generator produces a gif thumbnail for an mp4 video
+"""
```
JacksonMaxfield, line 61: No need to add this comments in the test file really. It should be clear by the API what is happening.

JacksonMaxfield (commented):
Lots of comments! Glad is working as is. Some recommendations on optimizations and test changes. Your tests are failing because you aren't using `resources_dir` (pytest is weird). And one of my tests is failing because of a bug I know about. If you could add this to your PR:

Under the `strategy` key, can you add `fail-fast: false`

https://github.com/CouncilDataProject/cdp-backend/blob/main/.github/workflows/check-pr.yml#L8

So that block should look like:
```yml
strategy:
  fail-fast: false
  matrix:
    python-version: [3.7, 3.8, 3.9]
    os: [ubuntu-latest, macOS-latest, windows-latest]
```

It will make it so all tests run regardless of if one of them fails.

Great work so far, I really appreciate it.

cdp_backend/utils/file_utils.py
```diff
@@ -169,6 +172,91 @@ def split_audio(
...
+    str: cover_name
+        The name of the thumbnail file: Always the same as the name of the video file,
+        only a 'png' replaces the 'mp4'
+    """
+
+    reader = imageio.get_reader(video_url)
+    png_path = ""
+    if reader.get_length() > 1:
+        png_path = (
+            f"{os.path.dirname(video_url)}/{session_content_hash}"
```
JacksonMaxfield, line 203: I would just save to current directory. No need to do video url or os dirname stuff imo.

just `png_path=f"{session_content_hash}-static-thumbnail.png"`

cdp_backend/utils/file_utils.py
```diff
@@ -169,6 +172,91 @@ def split_audio(
...
+            + "-static-thumbnail.png"
+        )
+
+    try:
+        frame_to_take = math.floor(reader.get_meta_data()["fps"] * seconds)
+        image = reader.get_data(frame_to_take)
+        imageio.imwrite(png_path, image)
```
JacksonMaxfield, line 210: nitpick: once you have the `image` variable, this imwrite becomes duplicated. Just move it out of the try except.

cdp_backend/utils/file_utils.py
```diff
@@ -169,6 +172,91 @@ def split_audio(
...
+    seconds: int
+        Determines after how many seconds a frame will be selected to produce the
+        thumbnail. The default is 30 seconds
+
+
+    Returns
+    -------
+    str: cover_name
+        The name of the thumbnail file: Always the same as the name of the video file,
+        only a 'png' replaces the 'mp4'
```
JacksonMaxfield, line 196: From code I think this docstring is now out-of-date.

cdp_backend/utils/file_utils.py
```diff
@@ -169,6 +172,91 @@ def split_audio(
...
+        image = reader.get_data(frame_to_take)
+        imageio.imwrite(png_path, image)
+    except:
+        image = reader.get_data(0)
+        imageio.imwrite(png_path, image)
+
+    return png_path
+
+
+def get_hover_thumbnail(
```
JacksonMaxfield, line 218: Same nitpick comments for this function.

cdp_backend/utils/file_utils.py
```diff
@@ -169,6 +172,91 @@ def split_audio(
...
 
+def get_static_thumbnail(
+    video_url: str, session_content_hash: str, seconds: int = 30
+) -> str:
+    """
+    A function that produces a png thumbnail image from an mp4 video file
+
+    Parameters
+    ----------
+    video_url: str
```
JacksonMaxfield, line 183: I would change this param to `video_path` as I think there is an assumption that the video will be local at the time this function runs.

cdp_backend/utils/file_utils.py
```diff
@@ -169,6 +172,91 @@ def split_audio(
...
+        png_path = (
+            f"{os.path.dirname(video_url)}/{session_content_hash}"
+            + "-static-thumbnail.png"
+        )
+
+    try:
+        frame_to_take = math.floor(reader.get_meta_data()["fps"] * seconds)
+        image = reader.get_data(frame_to_take)
+        imageio.imwrite(png_path, image)
+    except:
```
JacksonMaxfield, line 211: You should specify this as `KeyError` instead of just all exceptions. If something besides getting the `fps` key goes wrong we would want to know about it.

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -148,3 +152,62 @@ def test_split_audio(
 
         except Exception as e:
             raise e
+
+
+ending = "-static-thumbnail.png"
```
JacksonMaxfield, line 157: Can you move this variable into the test function. Have global vars in tests is somewhat rare.

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -148,3 +152,62 @@ def test_split_audio(
...
+        expected = str(resources_dir) + "/" + expected
+        result = file_utils.get_static_thumbnail(*args)
+        assert result == expected
+
+        assert Path(result).stat().st_size > 0
+
+        os.remove(result)
+
+
+ending = "-hover-thumbnail.gif"
```
JacksonMaxfield, line 185: Same here.

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -148,3 +152,62 @@ def test_split_audio(
...
+    ],
+)
+def test_static_thumbnail_generator(resources_dir: Path, args, expected: str):
+    args[0] = str(resources_dir) + "/" + args[0]
+
+    if type(expected) == type:
+        with pytest.raises(expected):
+            file_utils.get_static_thumbnail(*args)
+    else:
+        expected = str(resources_dir) + "/" + expected
```
JacksonMaxfield, line 176: Pathlib is actually great for this reason...

Instead of using the `"/"` character you can just do object division.

```python
expected = resources_dir / expected
```

the reasoning / benefit is that Windows uses a different path separator than Ubuntu and Mac. So that `"/"` will break if someone tried to run these tests on Windows. Plus it's just slightly cleaner imo.

JacksonMaxfield: You can see this failure in the test results on GitHub Actions.

cdp_backend/utils/file_utils.py
JacksonMaxfield: You can see this failure on the lint checks on GitHub Actions.

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -148,3 +152,62 @@ def test_split_audio(
...
+@pytest.mark.parametrize(
+    "args, expected",
+    [
+        ([example_file, "example"], "example" + ending),
+        ([example_file, "example2", 45], "example2" + ending),
+        (["fake.mp4", "example"], Exception),
+        (["fake_creds.json", "example"], Exception),
+    ],
+)
+def test_static_thumbnail_generator(resources_dir: Path, args, expected: str):
```
JacksonMaxfield, line 169: I would make all of the args explicit. `video_path: str, session_content_hash: str, seconds: int`.

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -148,3 +152,62 @@ def test_split_audio(
...
+ending = "-static-thumbnail.png"
+
+
+@pytest.mark.parametrize(
+    "args, expected",
+    [
+        ([example_file, "example"], "example" + ending),
+        ([example_file, "example2", 45], "example2" + ending),
+        (["fake.mp4", "example"], Exception),
+        (["fake_creds.json", "example"], Exception),
```
JacksonMaxfield, line 166: You can actually use a thing called `pytest.raises`.

Instead of this, you can do:
```python
pytest.param("fake.mp4", "example", marks=pytest.mark.raises(exception=KeyError))`
```

JacksonMaxfield: This allows your test code to not have type checking on expected. You can just run the function as you would expect it to run and if the provided error is raised, this parameter says "successful test"

JacksonMaxfield (commented):
Getting close, mostly nitpicks on the actual function but some larger test changes.

All tests (except for the webvtt_sr_model test) should pass. All linting should pass for sure.

cdp_backend/tests/utils/test_file_utils.py
Shak2000: I tried doing this earlier. However, I could not find a way to put the optional parameter `seconds` as `None`. I looked on different websites, and I found out that using `args` was recommended. I can either make the `seconds` parameter not optional, or I can omit tests that do not have this parameter.

JacksonMaxfield: It's optional in the non- test version. In the test version, just pass it as `None` as if it wasn't optional.

JacksonMaxfield: This test function def should be:

```python
@pytest.mark.parametrize(
    "video_path, session_content_hash, seconds, expected_save_path",
    [
        (example_file, "example", None, "example-static-thumbnail.png"),
        (example_file, "example2", 45, "example2-static-thumbnail.png"),
    ],
)
```

If you define all the arguments you can pass `None` as an argument.

cdp_backend/utils/file_utils.py
```diff
@@ -169,6 +172,86 @@ def split_audio(
...
+) -> str:
+    """
+    A function that produces a png thumbnail image from a video file
+
+    Parameters
+    ----------
+    video_path: str
+        The URL of the video from which the thumbnail will be produced
+    session_content_hash: str
+        The hash of the video from which the thumbnail will be produced
```
JacksonMaxfield, line 186: Can you change this to:

"The video content hash. This will be used in the produced image file's name."

cdp_backend/utils/file_utils.py
```diff
@@ -169,6 +172,86 @@ def split_audio(
...
+) -> str:
+    """
+    A function that produces a gif hover thumbnail from an mp4 video file
+
+    Parameters
+    ----------
+    video_path: str
+        The URL of the video from which the thumbnail will be produced
+    session_content_hash: str
+        The hash of the video from which the thumbnail will be produced
```
JacksonMaxfield, line 227: Same here.

JacksonMaxfield (commented):
Nice! Looking really good. We should get your local testing and linting figured out so that the last issues with CI can be resolved. Can you remind me what is happening locally that makes local tests and lint fail?

You should be able to run:

```bash
make clean
make build
```

And have nearly identical output as the GitHub Actions CI but locally.

JacksonMaxfield:
Oh nice! It's just the lint that is failing now (ignoring my failing test :sweat_smile:)

Just fix up the linting and my two minor comments and all good to go :+1: 

Awesome job!

Shak2000:
Resolves #65

JacksonMaxfield (approved):
Perfect thanks for all the changes! Merging now.

Issue #57, opened 2021-05-24
Add script for processing local file


JacksonMaxfield:
## Use Case
*Please provide a use case to help us understand your request in context*

To process the various one-off events that occur, for example, I just started processing the 2021 election events for Seattle.

## Solution
*Please describe your ideal solution*

I wrote a tiny script to do this in `cdptools` that takes a local file + some JSON file for the event / minutes details (the questions asked at the forum), uploads the video to Google storage then processes that uploaded file with the config JSON for event details.

In `cdp-backend` this is actually easier because all the ingestion models are seriabliable to JSON iirc. `dataclasses_json`.
So that would mean a user simply has to write the object in Python and serialize or just in pure JSON, then have a script accept the serialized ingestion model + the video path.

## Alternatives
*Please describe any alternatives you've considered, even if you've dismissed them*

PR #69, opened 2021-07-23
feature/event-gather-pipeline-transcription

### LINKED ISSUE #67: Make ingestion model optional in update_db_model function
## Description
- In `upload_db_model`, we allow `ingestion_model` to be an optional param. However, if the db model already exists and an ingestion model isn't passed, then a [uniqueness error is thrown](https://github.com/CouncilDataProject/cdp-backend/blob/4fb864281ec8d822b303a40b25aaf6a133ed88d1/cdp_backend/database/functions.py#L120-L133).
- There are some [places in the event gather pipeline](https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/event_gather_pipeline.py#L165-L174) where we call `upload_db_model_task` without a corresponding ingestion model because we don't really need an ingestion model. Th
- This is an unlikely edge case, but there could be a situation where an audio file is deleted from the file store, but the corresponding `file` db models are not. This could be happen if we're doing something like data cleanup or a backfill which causes inconsistencies.
- There are a few different ways to refactor this 
  -  We could fix this by changing the `upload_db_model` logic and removing the ingestion model `is not None` check, and changing `update_db_model` to take in `Union[Model, IngestionModel]` instead. It may also make more sense to have separate params for a db model vs ingestion model in `update_db_model` since the logic may differ a bit
  - We could also change `upload_db_model` to only accept a `db_model` instead of `ingestion_model`, and `update_db_model` to take in `existing_db_model` and `proposed_db_model`. This would probably make the both functions more simple, however would require us to create extra db models outside of those functions

## Expected Behavior
- Uniqueness error to not get triggered when `upload_db_model_task` is called with `None` for ingestion_model when an existing db model is present

JacksonMaxfield:

W.I.P #16 

Partially fixes #67 (@isaacna I know you had thoughts about how I was doing it on this branch)

Alright. This is a doozy.

This builds off of @isaacna's work with getting GCP transcription into the pipeline but there were quite a few changes I made. I know large PRs are the worst so I apologize in advance.

1. This adds both GCP transcript + closed caption "transcription" to the pipeline.
2. It adds existing audio and existing transcript checking.
3. It cleans up the pipeline in general to where each task is an encapsulated "serialization point" so that we can backfill large amounts of data without worrying about workers not having a file and such.
4. It adds config building and reading to the pipeline.
5. Cleans up some various functions and tests
6. Makes all ingestion models JSON serializable
7. Partially reverts the WebVTT -> CDP Transcript back to @tohuynh's version from `cdptools`

The pipeline diagrams didn't turn out as pretty as I would have hoped (I blame the fact that we are checking for existing transcript + audio) but here it is:

**Just GCP Transcription:**
image

**Both GCP Transcription and WebVTT Parsing:**
image

Still need to add tests for further down in the pipeline. Primarily transcript generation.

Thanks for contributing!


JacksonMaxfield:
Closing as I built off of this branch for #69

cdp_backend/bin/run_cdp_event_gather.py
```diff
@@ -33,66 +34,59 @@ def __parse(self) -> None:
...
-            "--google-credentials-file",
-            default=(Path(__file__).parent.parent.parent / "cdp-creds.json"),
-            type=Path,
-            dest="google_credentials_file",
-            help="Path to the Google Service Account Credentials JSON file.",
-        )
-        p.add_argument(
-            "-e",
-            "--get_events_function_path",
+            "config_file",
```
isaacna, line 37: Would we want to add a flag for this argument? Or is it convention to not need it since the config file is the main mandatory argument?

cdp_backend/database/functions.py
```diff
@@ -213,10 +208,24 @@ def create_session_from_ingestion_model(
     return db_session
 
 
-@task
-def create_file(name: str, uri: str) -> db_models.File:
+def create_file(uri: str) -> db_models.File:
     db_file = db_models.File()
-    db_file.name = name
+    db_file.name = uri.split("/")[-1]
```
isaacna, line 213: good catch here

cdp_backend/database/functions.py
```diff
@@ -108,25 +115,28 @@ def upload_db_model(
...
             f"Saved new {db_model.__class__.__name__} with document id={db_model.id}."
         )
     elif (
         len(uniqueness_validation.conflicting_models) == 1
         and ingestion_model is not None
     ):
         updated_db_model = update_db_model(
-            uniqueness_validation.conflicting_models[0],
-            ingestion_model,
+            db_model=uniqueness_validation.conflicting_models[0],
```
isaacna, line 132: Note to self: should still add fix for #67 here

cdp_backend/database/functions.py
```diff
@@ -108,25 +115,28 @@ def upload_db_model(
...
+            ingestion_model=ingestion_model,
+            transaction=transaction,
         )
 
         return updated_db_model
+    elif len(uniqueness_validation.conflicting_models) == 1 and exist_ok:
```
isaacna, line 138: Just to make sure I'm understanding the intent of the logic here, are we saying with the `exist_ok` param, we're fine with ignoring updates as well?

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -1,190 +1,730 @@
...
-                    session,
-                    creds_file=credentials_file,
+                    caption_confidence=config.caption_confidence,
+                )
+
+                # Generate thumbnails
+                (
+                    static_thumbnail_uri,
+                    hover_thumbnail_uri,
+                ) = get_video_and_generate_thumbnails(
```
isaacna, line 89: Nitpick, but do we need the parentheses here for assigning `static_thumbnail_uri` and `hover_thumbnail_uri`? I saw we didn't use that formatting above

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -1,190 +1,730 @@
...
     Returns
     -------
+    session_content_hash: str
+        The unique key (SHA256 hash of video content) for this session processing.
     audio_uri: str
-        The uri of the created audio file in the file store.
+        The URI to the uploaded audio file.
     """
     # Get just the video filename from the full uri
+    # TODO: Handle windows file splitting???
```
isaacna, line 143: Idk how likely it is that this would be ran off a windows machine, but still good to note

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -1,190 +1,730 @@
...
+    audio_uri = fs_functions.get_file_uri(
+        bucket=bucket,
+        filename=tmp_audio_filepath,
+        credentials_file=credentials_file,
     )
 
-    # If no existing audio uri, process video
-    with case(audio_uri, None):
+    # If no pre-existing audio, split
+    if audio_uri is None:
```
isaacna, line 166: So glad we don't have to mess with the prefect control flow here

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -1,190 +1,730 @@
...
+            bucket=bucket,
+            credentials_file=credentials_file,
+            session=session,
         )
 
-        # Upload files to database
-        uploaded_file_db_model = db_functions.upload_db_model_task(
-            audio_file_db_model, None, credentials_file
+    # Existing transcript
+    with case(transcript_uri, True):
```
isaacna, line 607: Shouldn't this be `case(transcript_exists, True)`?

cdp_backend/pipeline/pipeline_config.py
```diff
@@ -0,0 +1,73 @@
...
+            if self.gcs_bucket_name is not None:
+                bucket = self.gcs_bucket_name
+
+            else:
+                # Open the key to get the project id
+                with open(self.google_credentials_file, "r") as open_resource:
+                    creds = json.load(open_resource)
+                    project_id = creds["project_id"]
+
+                # Remove all files in bucket
```
isaacna, line 59: Did you mean to add code to remove files later?

isaacna (approved):
Thanks for doing this huge refactor! I think the new DAG is a big improvement. Even if it's still pretty complex, I think the way we encapsulate tasks will make debugging much easier and the pipeline less error prone in general. 

I left a few comments, but I don't think any of them are blocking/crucial to getting this merged.

cdp_backend/bin/run_cdp_event_gather.py
JacksonMaxfield: Which arg? The get events function path or the config file? By not having the `--` it makes the argument required and positional is all.

cdp_backend/database/functions.py
JacksonMaxfield: This is a great question.... The only time I ran into this / the reason I needed it was for the File DB model upload. Which doesn't need to be updated at all if it encounters a conflict. I don't know which models we would run into that need updating if they are found to exist...

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: This is just black. If the line is too long it will parenthesize and split line.

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: The only reason we would need to run on Windows is if a maintainer of a CDP instance wants to process a single event. See #57

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: Nice catch. I guess this was doing a falsy check?

cdp_backend/pipeline/pipeline_config.py
JacksonMaxfield: Oh lol. nope I just copied this code from the clean bucket bin script.

cdp_backend/bin/run_cdp_event_gather.py
isaacna: Yeah I meant `config_file`. I think you can force a flag to be required, but it wouldn't make it positional so nvm

cdp_backend/database/functions.py
isaacna: I was thinking about that whole edge case where Seattle channel archives videos and merges sessions to the old url. But I guess since we hash sessions by content we won't need to overwrite anything so this is fine. I do think we'd need to create another CRON job to "unlink" the other sessions from the event to handle redundant sessions from the merged case.

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Yeah I believe it would default to false in this case. According to [prefect docs](https://docs.prefect.io/api/latest/tasks/control_flow.html#case) it checks equality of the task return val and the second param.

cdp_backend/database/functions.py
isaacna: Now that I think about it, would there any case where we need to update the db models instead of returning the existing ones?

JacksonMaxfield: Until we get a bug report of it. I say this is fine??? Maybe?

isaacna: Hm yeah idk, I'm fine keeping the updating code or just returning the existing model. Even if the update happens (which may not be likely) I think that'll be ok

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Ah gotcha good point, didn't think about that use case

PR #70, opened 2021-07-27
feature/handle-all-ingestion-metadata

JacksonMaxfield:

W.I.P. #16 (Just waiting on #65 and we can close!! :tada: )

Handles all the metadata / ingestion model processing and uploading. Basically just one giant mapping from ingestion model to database model and various extra file uploading we do.

You are either going to hate this, love this, or not care. But strongly in one of those three. :upside_down_face: 

Open to feedback if you hate it thought :slightly_smiling_face: 

Thanks for contributing!


cdp_backend/database/models.py
```diff
@@ -53,12 +58,15 @@ class Person(Model):
...
     email = fields.TextField(validator=validators.email_is_valid)
-    phone = fields.NumberField()
+    phone = fields.TextField()
     website = fields.TextField(validator=validators.resource_exists)
     picture_ref = fields.ReferenceField(File)
     is_active = fields.BooleanField(required=True)
     external_source_id = fields.TextField()
 
+    class Meta:
+        ignore_none_field = False
```
isaacna, line 68: Not sure if you intended the `Meta` inner class to be similar across all models, but maybe we could do something like the following for all the db models. I think having them inherit would reduce code duplication if we expect metadata to be similar
```
class MetaModel:
    class Meta:
        ignore_none_field = False

class File(Model, MetaModel):
```

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -626,8 +636,30 @@ def generate_transcript(
...
 def get_video_and_generate_thumbnails(
-    session_content_hash: str, video_uri: str, bucket: str, credentials_file: str
+    session_content_hash: str,
+    video_uri: str,
+    event: EventIngestionModel,
+    bucket: str,
+    credentials_file: str,
 ) -> Tuple[str, str]:
+    if event.static_thumbnail_uri is None:
+        # Generate new
```
isaacna, line 646: This is placeholder code for Shak's stuff right?

cdp_backend/tests/database/test_functions.py
```diff
@@ -223,66 +160,29 @@ def test_upload_db_model(
                     assert_db_models_equality(expected, actual_uploaded_model, True)
 
 
-def test_create_file() -> None:
-    db_file = db_functions.create_file("ex://name.ext")
+###############################################################################
 
-    assert example_file.name == db_file.name
-    assert example_file.uri == db_file.uri
+# Only test functions that do something besides parameter unpacking and assigning
```
isaacna, line 165: Good call lol

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -653,31 +685,262 @@ def compile_session_processing_result(
...
 def store_event_processing_results(
     event: EventIngestionModel,
     session_processing_results: List[SessionProcessingResult],
     credentials_file: str,
+    bucket: str,
 ) -> None:
+    # TODO: check metadata before pipeline runs to avoid the many try excepts
+
+    # Get session with earliest datetime for any event level metadata
+    first_session = min(event.sessions, key=attrgetter("session_datetime"))
```
isaacna, line 867: Could we also use `session_index` instead? Or if we can just sort by `session_datetime` do we even need `session_index`?

isaacna (approved):
Looks good, can't say I looked at every line of code for the `event_gather_pipeline` changes, but the db model creation and upload pattern looked good.

One thought I had was whether you think it's worth it to break up the pipeline file further as it's getting pretty large.

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: Yep!

JacksonMaxfield:
> Looks good, can't say I looked at every line of code for the `event_gather_pipeline` changes, but the db model creation and upload pattern looked good.
> 
> One thought I had was whether you think it's worth it to break up the pipeline file further as it's getting pretty large.

I think overtime we will. I think there will be pipelines that do similar things that we will begin moving tasks to a general tasks file or something similar. I agree though, it's getting larger and larger and part of that we can't really help but parts we can over time.

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: I will use index. Iirc we discussed this way back when making the DB schema and we have both because a person may know the session ordering but not the exactly start times of each session. i.e. they just use the same event datetime for all the sessions even though one happened in the morning and one happened in the afternoon.

cdp_backend/database/models.py
JacksonMaxfield: I tried to do this and it results in really weird behavior around the class attrs.
I.e.

```python
File.Example().model.collection
```

just fails because the created example file object doesn't have a model attr???

Going to leave it as it is for now. I was annoyed by it too though.

PR #72, opened 2021-08-04
bugfix/matter-status-and-external-source

JacksonMaxfield:

Pipeline bugfixes found while testing: https://github.com/CouncilDataProject/cdp-scrapers/pull/1

While the expectation is that `external_source_id` is a string, lets just cast it to a string regardless.
Make `matter.result_status` ingestion optional as there are matters that are not voted on and have no status. Primarily "Information Items (Inf)" such as presentations and such.

Thanks for contributing!

### DIFF
--- a/cdp_backend/database/functions.py
+++ b/cdp_backend/database/functions.py
@@ -179,7 +179,7 @@ def create_body(
     # Optional fields
     db_body.end_datetime = body.end_datetime
     db_body.description = _strip_field(body.description)
-    db_body.external_source_id = body.external_source_id
+    db_body.external_source_id = str(body.external_source_id)
 
     return db_body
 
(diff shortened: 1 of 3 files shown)


tohuynh (approved):
Looks good.

I think we may need to add another field to MatterStatus to describe matter status. Currently, on the legislation details page, we would be able to show the votes and decision for a time point in the legislation's history, but not what the timepoint is about(e.g. hold bill, send bill to full council).

JacksonMaxfield:
> I think we may need to add another field to MatterStatus to describe matter status. Currently, on the legislation details page, we would be able to show the votes and decision for a time point in the legislation's history, but not what the timepoint is about(e.g. hold bill, send bill to full council).

Yea thought about that too. Fortunately shouldn't be too hard to change. And maybe something to look into when we start actually working on the legislation pages.

isaacna (approved):

PR #74, opened 2021-08-08
feature/deterministic-document-keys

JacksonMaxfield:

This changes how we are storing data in the database. The primary focus of these changes are to make doc ids deterministic by hashing the primary values of the doc. This means we can simply use the `upsert` command instead of doing a query then insert to determine model uniqueness. We just don't need to care about uniqueness because the model is always deterministic. It also "gracefully" handles race conditions in parallel processing.

Removes a ton of tests and random functions because our upload is much simpler now.

Thanks for contributing!


isaacna (approved):
This is a huge improvement to the even gather pipeline and how we handle db uploads in general! I think making the doc keys deterministic prevents pretty much covered the biggest potential space for error in our pipeline

PR #78, opened 2021-08-12
Adding thumbnail generation to the pipeline

### LINKED ISSUE #65: Generate thumbnail PNG and GIF for each video

Shak2000:
### Link to Relevant Issue

Resolves #65 

### Description of Changes

The pipeline will now be able to create a static and hover thumbnail for any video file that it receives as input.


JacksonMaxfield (commented):
Biggest comment is that if the static or hover thumbnail are provided with the event details, you should download them then upload them to the file store and return _our_ copy of them.

I think you also said tests were coming so I will hold off any other comments until then.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -665,27 +666,67 @@ def get_video_and_generate_thumbnails(
...
+        hover_thumbnail_file = file_utils.get_hover_thumbnail(
+            tmp_video_path, session_content_hash
+        )
+
+        hover_thumbnail_url = fs_functions.upload_file(
+            credentials_file=credentials_file,
+            bucket=bucket,
+            filepath=hover_thumbnail_file,
+        )
+        fs_functions.remove_local_file(hover_thumbnail_file)
```
JacksonMaxfield, line 723: This all looks fine but you need to download the existing static and hover thumbnail files if provided as a part of the event model.

cdp_backend/tests/pipeline/test_event_gather_pipeline.py
```diff
@@ -103,6 +103,67 @@ def test_get_video_and_split_audio(
...
+            f"fake://{VIDEO_CONTENT_HASH}-hover-thumbnail.gif",
+            VIDEO_CONTENT_HASH,
+            EXAMPLE_MINIMAL_EVENT,
+            False,
+        ),
+    ],
+)
+def test_get_video_and_generate_thumbnails(
+    mock_upload_file: MagicMock,
+    # mock_get_file_uri: MagicMock,
```
JacksonMaxfield, line 133: Remove commented out parameter

cdp_backend/tests/pipeline/test_event_gather_pipeline.py
```diff
@@ -103,6 +103,67 @@ def test_get_video_and_split_audio(
...
+    is_static: bool,
+    example_video: Path,
+) -> None:
+    # Since mock_upload_file only allows for one return value and since the real
+    # thumbnail generator calls upload_file twice, it is necessary to test each
+    # thumbnail generation process separately
+    if is_static:
+        mock_upload_file.return_value = example_static_thumbnail_url
+    else:
+        mock_upload_file.return_value = example_hover_thumbnail_url
```
JacksonMaxfield, line 147: You should check out [mock.side_effect](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect) as it allows for exactly this behavior of multiple calls to the same function. You can provide a list of returns and it will iterate over the list each time the mocked function is called.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -692,35 +692,43 @@ def get_video_and_generate_thumbnails(
     """
     tmp_video_path = file_utils.resource_copy(video_uri)
 
+    static_thumbnail_file = ""
```
JacksonMaxfield, line 695: Just curious, are these here for typing reasons?

JacksonMaxfield (approved):
Nice! One minor question but happy to merge whenever!

JacksonMaxfield:
Oh and ignore the coverage patch check.

cdp_backend/pipeline/event_gather_pipeline.py
Shak2000: I thought it was necessary in order to give the variables the right scope (confusion with Java).

Shak2000: I just pushed a commit to remove it.

JacksonMaxfield: Ahhhh gotcha. Waiting for tests to pass and will then merge!

PR #79, opened 2021-08-13
admin/improve-dev-infra-docs

isaacna:
### Link to Relevant Issue

This pull request resolves #

### Description of Changes

- Improving the docs in `dev-infrastructure` to be a little more specific

### DIFF
--- a/dev-infrastructure/README.md
+++ b/dev-infrastructure/README.md
@@ -59,18 +59,18 @@ following commands to setup your local machine with credentials to both services
 need to run these scripts in a py37 environment.
 
 ```bash
-cd ~/cdp-backend/dev-infrastructure
+cd cdp-backend/dev-infrastructure
 make login
 make init project={project-name}
 make gen-key project={project-name}
 make build
 ```
 
-After generating the key, name your `key` file in `~/cdp-backend/.keys` to `cdp-dev.json`. In case you have many keys, note that by default, the random and minimal event pipelines use the key named `cdp-dev.json`.
+After generating the key, name your `key` file in `cdp-backend/.keys` to `cdp-dev.json`. In case you have many keys, note that by default, the random and minimal event pipelines use the key named `cdp-dev.json`.
 
 
 ## Infrastructure Management Commands
-All these commands should be run from `~/cdp-backend/dev-infrastructure`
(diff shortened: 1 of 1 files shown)


dev-infrastructure/README.md
```diff
@@ -168,3 +175,5 @@ base directory of this repository and running the following:
     ```bash
     make run-min-event-pipeline
     ```
+
+Where `~` is the directory where this repo lives.
```
JacksonMaxfield, line 179: My nitpick here is just that `~` is a unix operator so doesn't work in Windows and it also means "home dir" which isn't where I store my code stuff.

I would just remove all the `~/` stuff throughout the changes.

Anywhere I see a `~/` just remove it.

JacksonMaxfield (approved):
All looks good! Hope you found the process easy enough.

One comment but otherwise happy to merge

dev-infrastructure/README.md
```diff
@@ -59,16 +59,18 @@ following commands to setup your local machine with credentials to both services
...
+make build
 ```
 
-Rename the `key` file in `../keys` file `cdp-dev.json`. 
+After generating the key, name your `key` file in `~/cdp-backend/.keys` to `cdp-dev.json`. In case you have many keys, note that by default, the random and minimal event pipelines use the key named `cdp-dev.json`.
 
 
-## Infrastrure Management Commands
+## Infrastructure Management Commands
+All these commands should be run from `~/cdp-backend/dev-infrastructure`
```
hawkticehurst, line 73: If we're going for precision in language I would probably restate this as:

> All of these commands should be run from within the `~/cdp-backend/dev-infrastructure` directory.

hawkticehurst (approved):
Beyond one comment, looks good to me!

dev-infrastructure/README.md
isaacna: Yeah, I was thinking about putting something like `{ROOT}/cdp-backend` since I wasn't sure how to describe it, but removing is still clear enough

dev-infrastructure/README.md
isaacna: That sounds better, will make that change!

PR #80, opened 2021-08-13
feature/event-index-pipeline

### LINKED ISSUE #23: Add EventIndexPipeline
## Use Case

Create a searchable index for the events gathered and upload it to Firestore.

## Solution

Migrate over and clean up the work done in [cdptools@feature/update-index-pipeline](https://github.com/CouncilDataProject/cdptools/tree/feature/update-index-pipeline)

Additionally, see if this can entirely be done on GH Actions or if we need to use compute engine. Keeping the cost low would be great.

## Alternatives

### LINKED ISSUE #24: Add Database Models for Indexed Event Terms
## Use Case

Need collections to store the created indexed terms for events

## Solution

Complete the work laid out in [cdptools #79](https://github.com/CouncilDataProject/cdptools/issues/79) some of this may be incomplete, look to [cdptools@feature/update-index-pipeline](https://github.com/CouncilDataProject/cdptools/tree/feature/update-index-pipeline) for most up-to-date.

## Alternatives

JacksonMaxfield:
### Link to Relevant Issue

Resolves #23
Resolves #24

### Description of Changes

There are a lot of files touched and some of them are minor bugfixes I found while working on this + admin updates.

**Core additions:**
The event index pipeline has been fully added to the library. It utilizes [TF*IDF](https://en.wikipedia.org/wiki/Tf%E2%80%93idf) as the base method for scoring n-grams. This is the same underlying algorithm as `cdptools` however there are many upgrades.

1. We now include both the "pure tfidf relevance" scores in the database model _and_ the score weighted by the timedelta from index processing - event datetime. `(tf * idf) / log(((pt - et) / 30) + 2)` where `tf` = term frequency of the term in the document, `idf` is the inverse document frequency, `pt - et` is the difference in days between the current datetime of processing and the event datetime. The ` / 30` term converts the days difference into float of months and the `+ 2` ensures no division by zero.

2. We can now generate `n-grams` (multiple term indexed items). This should drastrically increase the _specificity_ of query results. I.e. if someone searches "housing affordability" the indexed event gram may have "hous afford" as the stemmed gram to get the value. In doing so, we can run this same pipeline with a different `n_gram` integer value to generate as high of an n gram size as we want.

3. The pipeline is _much_ faster. Credit to `fireo` for making database interactions easy and our own database models, and our transcript model, the actual indexing pipeline is pretty simple. Additionally, because fireo has a great interface for batching uploads, we can chunk our index upload into batches instead of uploading a single term at a time. Firebase puts a limit of 500 items max in a single batch. I put the batch size at 400 just for safety but we can increase to shave a few more seconds off the pipeline.

4. A proper search utility. You can quickly and easily run a search against both a local index and remote database index with a single command to compare during development. I.e. if we want to change the weighting or how we index terms, we can generate and store an index locally and run the search utility to see the effect of the change. And you can sort by either datetime weighted or pure relevance to see the difference that makes. **Note: the remote search is _slow_ -- ~20 seconds. This is because of how we have `fireo` setup. Everytime we pull a database model that has a reference field it also pulls the linked document in full, recursively. So in the case of this search utility, that means we are pulling three documents for every matching `IndexedEventGram` (`IndexedEventGram` -> `Event` -> `Body` [and optionally two `File` documents for the event thumbnails]). This behavior will not happen on the web application side. This is an effect of `fireo`.**

5. Context spans! An ask from the UX team way back when was to include _some_ sort of context as to why the events were being returned in the order they were in. As a part of the indexing pipeline I grab the surrounding 14 words (including the n-gram itself, this means the context can range from 15 - 18 terms).

**Bugfixes and Admin:**
1. gitignore additions
2. editorconfig changes that result in github action changes
3. logging additions
4. standardizing bin script "single char parameters" (`-p` instead of `--p`)
5. a bugfix for the webvtt sr model that caused the model to miss a lot of new speaker patterns because of the `$` regex "end of sentence" anchor

<summary><b>Example of Search Utility Against Last Month of Seattle Events:</b></summary>
<details>

```
(cdp-backend) maxfield@maxfield-ubuntu:~/active/cdp/cdp-backend$ search_cdp_events --query "residential zoning and housing affordability" --sort_by datetime_weighted_relevance
[INFO: search_cdp_events: 130 2021-08-13 02:19:13,286] Running search against remote index...
Match 0: 52cd769e7209 (datetime: 2021-08-11 09:30:00+00:00)
Match pure tf-idf relevance: 46.299532250147905
Match datetime weighted relevance: 3.8574075686051392
Match contained grams: ['residential', 'zones', 'housing', 'Affordability', 'residential zones', 'housing Affordability']
Match keywords: ['StreetLevel', 'office spaces', 'storefronts', 'CB 120121', 'downtown']
Match context: ... establish alternative building establishments in downtown mixed residential zones.
--------------------------------------------------------------------------------
Match 1: 91f2305e8455 (datetime: 2021-07-28 09:30:00+00:00)
Match pure tf-idf relevance: 49.55546123356822
Match datetime weighted relevance: 3.5065287447010207
Match contained grams: ['residential', 'zones', 'housing', 'affordable', 'residential zone']
Match keywords: ['single family', 'Council bill 120108', 'neighborhood residential', 'tree', 'Chair Strauss']
Match context: ... the up--to update the name of single family zones to neighborhood residential.
--------------------------------------------------------------------------------
Match 2: ae0732248895 (datetime: 2021-07-14 09:30:00+00:00)
Match pure tf-idf relevance: 45.626683587700256
Match datetime weighted relevance: 3.0903504924026155
Match contained grams: ['residential', 'zones', 'housing', 'affordable', 'residential zones', 'zoning housing', 'housing affordable']
Match keywords: ['tree', 'canopy', 'comprehensive plan', 'tree ordinance', 'tree canopy']
Match context: ... it contributed to the trees in our residential zones, we cannot build apartments.
--------------------------------------------------------------------------------
Match 3: d4ef9bbdd6dc (datetime: 2021-08-02 14:00:00+00:00)
Match pure tf-idf relevance: 13.324214889473863
Match datetime weighted relevance: 0.9699411874419692
Match contained grams: ['residential', 'zoning', 'house', 'affordable', 'residential zoning', 'housing affordable']
Match keywords: ['new hope', 'Pederson', 'Maria', 'church', 'August 9th']
Match context: ... its act of complicity and to reinvest in affordable housing.
--------------------------------------------------------------------------------
[INFO: search_cdp_events: 220 2021-08-13 02:19:33,063] Completed remote search in: 0:00:19.774692
[INFO: search_cdp_events: 229 2021-08-13 02:19:33,063] Running search against local index...
[INFO: search_cdp_events: 241 2021-08-13 02:19:33,074] No local index found.
```

</details>


JacksonMaxfield:
I am heading to bed but will add tests tomorrow!

JacksonMaxfield:
I just realized a better date for weighting the `datetime_weighted_value` for the indexed term. Instead of epoch time (1970, 1,1), I should use the earliest event datetime currently stored in the database at processing time. To get a more relevant log scale out of the dates.

tohuynh:
> I just realized a better date for weighting the `datetime_weighted_value` for the indexed term. Instead of epoch time (1970, 1,1), I should use the earliest event datetime currently stored in the database at processing time. To get a more relevant log scale out of the dates.

Are you trying to boost the overall score for more recent events? If yes, then you don't really need to do this since it's the log of the timedelta anyways.

JacksonMaxfield:
> > I just realized a better date for weighting the `datetime_weighted_value` for the indexed term. Instead of epoch time (1970, 1,1), I should use the earliest event datetime currently stored in the database at processing time. To get a more relevant log scale out of the dates.
> 
> Are you trying to boost the overall score for more recent events? If yes, then you don't really need to do this since it's the log of the timedelta anyways.

@tohuynh yep totally agree. I just tested it out with two variants:
```python
UTCNOW = datetime.utcnow()
UTCNOW = pytz.timezone("UTC").localize(UTCNOW)
n_grams["datetime_weighted_tfidf"] = n_grams.apply(
    lambda row: row.tfidf / math.log(((UTCNOW - row.event_datetime).days / 30) + 2),
    axis=1,
)
```

and
```python
UTCNOW = datetime.utcnow()
UTCNOW = pytz.timezone("UTC").localize(UTCNOW)
n_grams["datetime_weighted_tfidf"] = n_grams.apply(
    lambda row: row.tfidf / math.log((UTCNOW - row.event_datetime).days + 2),
    axis=1,
)
```

Both work well, I somewhat side with the first one as it introduces a term the `- 30` that basically smoothens out the values a bit more and in my head says "anything in the last month should count as _more_ relevant than purely the tfidf score.

We could change that `- 30` term to be something else, maybe "anything in the last 14 days should be _more relevant_ than the purely tfidf score". It's just a partition basically. Thoughts?

(the `+ 2` term at the end just makes it so we don't encounter divide by zero errors)

tohuynh:
That makes sense. You changing the timedelta unit to month (but in floating point value instead of integer).

30 days seems fine.

JacksonMaxfield:
Already tests added. Review whenever.

cdp_backend/pipeline/event_index_pipeline.py
```diff
@@ -0,0 +1,518 @@
...
+    documents as this will additionally pull all referenced Event, Session, Body,
+    and File database models.
+    """
+    fireo.connection(from_file=credentials_file)
+
+    # Fetch all transcripts
+    # This comes with references to Session and File models
+    # Session models come with reference to Event model
+    # Event models come with reference to Body model
+    return list(db_models.Transcript.collection.fetch(limit=int(1e9)))
```
isaacna, line 77: Isn't this value pretty high? Just thinking that if FireO fetches reference models recursively, latency concerns aside, couldn't this lead to a stack overflow pretty easily?

cdp_backend/pipeline/event_index_pipeline.py
```diff
@@ -0,0 +1,518 @@
...
+    event_transcripts: Dict[str, EventTranscripts] = {}
+    for transcript in transcripts:
+        # Add new event
+        if transcript.session_ref.event_ref.id not in event_transcripts:
+            event_transcripts[transcript.session_ref.event_ref.id] = EventTranscripts(
+                event=transcript.session_ref.event_ref,
+                transcripts=[transcript],
+            )
+
+        # Update existing eventtranscripts object
```
isaacna, line 141: nitpick: `event_transcripts`

isaacna:
> We could change that `- 30` term to be something else, maybe "anything in the last 14 days should be _more relevant_ than the purely tfidf score". It's just a partition basically. Thoughts?
> 
> (the `+ 2` term at the end just makes it so we don't encounter divide by zero errors)

I was actually thinking that it might be preferable to not use the `/ 30`? If we consider that it's basically converting to months, that might not put enough emphasis on recent events. Like if council meetings happen (weekly?) on recurring topics, I think doing the math based on # days would add more weight to recency. Especially with adding the `+ 2` to avoid dividing by 0, I think weighting by months not be enough since we take the log of that whole value.

But also I'm v rusty at math so take that with a grain of salt lol

isaacna (approved):
Looks good! I'll admit I didn't look at all the lines of code for implementing the n grams, tfidf, etc. very carefully for correctness. My comments were mostly just about our strategy on weighting and the db query limit

cdp_backend/pipeline/event_index_pipeline.py
JacksonMaxfield: It's _dangerous_ but not that dangerous. In the production cdp-seattle v2 instance we have right now we have a total of 400 transcripts. That said. I can break this up into chunks.

JacksonMaxfield:
> > We could change that `- 30` term to be something else, maybe "anything in the last 14 days should be _more relevant_ than the purely tfidf score". It's just a partition basically. Thoughts?
> > (the `+ 2` term at the end just makes it so we don't encounter divide by zero errors)
> 
> I was actually thinking that it might be preferable to not use the `/ 30`? If we consider that it's basically converting to months, that might not put enough emphasis on recent events. Like if council meetings happen (weekly?) on recurring topics, I think doing the math based on # days would add more weight to recency. Especially with adding the `+ 2` to avoid dividing by 0, I think weighting by months not be enough since we take the log of that whole value.
> 
> But also I'm v rusty at math so take that with a grain of salt lol

So yea. I actually spent the time to plot out how the relevance decays and exactly how To described it above, that term is just change the unit of decay to a month but with the precision of a float. So instead of integer days, it becomes float months. And we can simply change this to a different unit of decay. What I saw in plotting the alg for different timedeltas was that anything in the last month was given a higher datetime weighted relevance than pure relevance and anything more than a month old was given a lower datetime weighted relevance than pure relevance. I personally think that is reasonable.

What is the "unit of time for city council meetings to be relevant" and I think it is on the month scale not the week or day scale.

cdp_backend/pipeline/event_index_pipeline.py
JacksonMaxfield: On the stack overflow issue, no that would be crazy to reach... these are the Transcript db models not the transcript file models. These are like KB's large.

JacksonMaxfield:
I made the "datetime_weighting_days_decay" a parameter for the whole pipeline and its store in the index pipeline config so that people can easily change the relevance weighting.

isaacna:
Gotcha in that case then yeah I think the months scale is the move. And yeah the optional decay param is a good call too

PR #83, opened 2021-08-14
feature/pass-datetimes-to-get-events-func

JacksonMaxfield:
### Description of Changes

Working on cookiecutter / integration, I rememberd that GitHub Actions has an event called [Workflow Dispatch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch) which allows for running pipelines manually through the GitHub UI: [Docs](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) and figured that it would be a great way to allow for backfilling data on the github action runner instead of a persons local machine.

So this adds the `from` and `to` arguments to the bin script and the `from_dt` and `to_dt` arguments to the pipeline constructor itself and passes the values down to the `get_events` function that was loaded. Minor change to `get_events` spec but I think a worthwhile one.

### DIFF
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -2,10 +2,10 @@
 # -*- coding: utf-8 -*-
 
 import logging
-from datetime import timedelta
+from datetime import datetime, timedelta
 from importlib import import_module
 from operator import attrgetter
-from typing import Any, Callable, Dict, List, NamedTuple, Optional, Set, Tuple
+from typing import Any, Callable, Dict, List, NamedTuple, Optional, Set, Tuple, Union
 
 from fireo.fields.errors import FieldValidationFailed, InvalidFieldType, RequiredField
 from fsspec.core import url_to_fs
@@ -48,7 +48,11 @@ def import_get_events_func(func_path: str) -> Callable:
     return getattr(mod, func_name)
 
 
-def create_event_gather_flow(config: EventGatherPipelineConfig) -> Flow:
+def create_event_gather_flow(
+    config: EventGatherPipelineConfig,
+    from_dt: Optional[Union[str, datetime]] = None,
+    to_dt: Optional[Union[str, datetime]] = None,
+) -> Flow:
     """
     Provided a function to gather new event information, create the Prefect Flow object
     to preview, run, or visualize.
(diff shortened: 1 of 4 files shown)


cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -66,10 +78,35 @@ def create_event_gather_flow(config: EventGatherPipelineConfig) -> Flow:
     # Load get_events_func
     get_events_func = import_get_events_func(config.get_events_function_path)
 
+    # Handle from datetime
+    if isinstance(from_dt, str) and len(from_dt) != 0:
+        from_datetime = datetime.fromisoformat(from_dt)
+    elif isinstance(from_dt, datetime):
+        from_datetime = from_dt
+    else:
+        from_datetime = datetime.utcnow() - timedelta(days=2)
```
isaacna, line 87: Is 2 days just an arbitrary value? Would it make sense to match the default `from_dt` with the frequency of how often we run the pipeline [in cookiecutter](https://github.com/CouncilDataProject/cookiecutter-cdp-deployment/blob/main/%7B%7B%20cookiecutter.hosting_github_repo_name%20%7D%7D/.github/workflows/event-processing-pipeline.yml#L12)?

JacksonMaxfield: It's sorta artibitrary.

What I have seen in the past is that events commonly get pushed to legistar in Seattle within a couple of hours but then over the course of two days they are updated with more info.

Definitely not matching the CRON schedule for running the pipeline though. Because of the above issues if we only collected events every 4 hours for the last 4 hours we would miss a lot.

JacksonMaxfield: It's sorta artibitrary.

What I have seen in the past is that events commonly get pushed to legistar in Seattle within a couple of hours but then over the course of two days they are updated with more info.

Definitely not matching the CRON schedule for running the pipeline though. Because of the above issues if we only collected events every 4 hours for the last 4 hours we would miss a lot.

JacksonMaxfield: Eh I will make it a config parameter haha

isaacna (approved):
Looks good! Also, I created [this issue](https://github.com/CouncilDataProject/cookiecutter-cdp-deployment/issues/40) in cookiecutter to update the github action once this is merged:

JacksonMaxfield:
> Looks good! Also, I created [this issue](https://github.com/CouncilDataProject/cookiecutter-cdp-deployment/issues/40) in cookiecutter to update the github action once this is merged:

Thanks! Already working on it so will close that issue when my PR is ready to go.

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Ah gotcha I didn't know that they did that with legistar, I think 2 days makes sense then

isaacna: Yeah I actually think making it a config param rather than script arg is better (and makes it easier to update in the cookiecutter as well)

JacksonMaxfield: Uhhhh about to push. You will see what I mean.

JacksonMaxfield: Pushed, see the config option added.

isaacna:
Saw the updates, yeah I think having time delta as a config option for the github workflows makes more sense than from and to dates

PR #89, opened 2021-08-17
feature/process-local-file

### LINKED ISSUE #57: Add script for processing local file

isaacna:
### Link to Relevant Issue

This pull request resolves #57 

### Description of Changes

Adds a script for processing a local video file. `event_details_file` is just an `EventIngestionModel` in JSON format.

Example run:
```
process_local_file --local_video_file wombo_combo.mp4 --event_details_file event-details.json --event_gather_config_file event-gather-config.json
```

I also made the `resource_copy` function create the file in a temporary directory if `dst` is not provided. I did this because otherwise the `resource_copy` will copy into the repo directory. This usually isn't a problem, but for example if the video file provided in the script lives in `cdp_backend`, then we'll get a `FileExistsError` during resource copy and the pipeline will error out.

### DIFF
--- a/cdp_backend/bin/process_local_file.py
+++ b/cdp_backend/bin/process_local_file.py
@@ -6,10 +6,15 @@ import logging
 import sys
 import traceback
 from pathlib import Path
+from datetime import datetime
 
 import fireo
 
 from cdp_backend.database import DATABASE_MODELS
+from cdp_backend.file_store.functions import upload_file
+from cdp_backend.pipeline.pipeline_config import EventGatherPipelineConfig
+from cdp_backend.pipeline.ingestion_models import EventIngestionModel, Session
+from cdp_backend.pipeline import event_gather_pipeline as pipeline
 
 ###############################################################################
 
@@ -28,16 +33,80 @@ class Args(argparse.Namespace):
 
     def __parse(self) -> None:
         p = argparse.ArgumentParser(
-            prog="clean_cdp_database",
-            description="Delete all collections in a CDP database.",
+            prog="process_local_file",
+            description="Process local video file into the event pipeline.",
         )
         p.add_argument(
-            "google_credentials_file",
+            "--local_video_file",
             type=Path,
-            help="Path to Google service account JSON key.",
+            help="Path to the video file to process into the event pipeline.",
+        )
+        p.add_argument(
+            "--event_details_file",
+            type=Path,
+            help="Path to the JSON file with event details.",
+        )
+        p.add_argument(
+            "--event_gather_config_file",
+            type=Path,
+            help="Path to the JSON file with event gather pipeline configuration.",
         )
         p.parse_args(namespace=self)
 
 
+def main() -> None:
+    try: 
+        args = Args()
+
+        # Read pipeline config
+        with open(args.event_gather_config_file, "r") as open_resource:
+            config = EventGatherPipelineConfig.from_json(  # type: ignore
+                open_resource.read()
+            )
+
+        # Upload video file to file store
+        video_uri = upload_file(
+            credentials_file=config.google_credentials_file,
+            bucket=config.validated_gcs_bucket_name,
+            filepath=str(args.local_video_file),
+        )
+
+        # Convert event details file to EventIngestionModel
+        with open(args.event_details_file, "r") as open_resource:
+            ingestion_model = EventIngestionModel.from_json(  # type: ignore
+                open_resource.read()
+            )
+
+            # Create session from video_uri
+            session = Session(
+                # TODO: Should we instead take in datetime as an input?
+                session_datetime=datetime.utcnow(),
+                video_uri=video_uri,
+                session_index=0,
+            )
+
+            # Attach session to event
+            ingestion_model.sessions = [session]
+
+        # Create event gather pipeline flow
+        flow = pipeline.create_event_gather_flow(
+            config=config,
+            prefetched_events=[ingestion_model]
+        )
+
+        # Run flow
+        flow.run()
+
+    except Exception as e:
+        log.error("=============================================")
+        log.error("\n\n" + traceback.format_exc())
+        log.error("=============================================")
+        log.error("\n\n" + str(e) + "\n")
+        log.error("=============================================")
+        sys.exit(1)
+
+
 ###############################################################################
 
+if __name__ == "__main__":
+    main()
--- a/cdp_backend/bin/process_local_file.py
+++ b/cdp_backend/bin/process_local_file.py
@@ -73,7 +73,7 @@ def main() -> None:
 
         # Create event gather pipeline flow
         flow = pipeline.create_event_gather_flow(
-            config=config, prefetched_events=[ingestion_model]
+            config=config, prefetched_events=[ingestion_model], from_local=True
         )
 
         # Run flow
--- a/cdp_backend/bin/process_local_file.py
+++ b/cdp_backend/bin/process_local_file.py
@@ -5,16 +5,13 @@ import argparse
 import logging
 import sys
 import traceback
-from pathlib import Path
 from datetime import datetime
+from pathlib import Path
 
-import fireo
-
-from cdp_backend.database import DATABASE_MODELS
 from cdp_backend.file_store.functions import upload_file
-from cdp_backend.pipeline.pipeline_config import EventGatherPipelineConfig
-from cdp_backend.pipeline.ingestion_models import EventIngestionModel, Session
 from cdp_backend.pipeline import event_gather_pipeline as pipeline
+from cdp_backend.pipeline.ingestion_models import EventIngestionModel, Session
+from cdp_backend.pipeline.pipeline_config import EventGatherPipelineConfig
 
 ###############################################################################
 
@@ -55,7 +52,7 @@ class Args(argparse.Namespace):
 
 
 def main() -> None:
-    try: 
+    try:
         args = Args()
 
         # Read pipeline config
@@ -90,8 +87,7 @@ def main() -> None:
 
         # Create event gather pipeline flow
         flow = pipeline.create_event_gather_flow(
-            config=config,
-            prefetched_events=[ingestion_model]
+            config=config, prefetched_events=[ingestion_model]
         )
 
         # Run flow
--- a/cdp_backend/bin/process_local_file.py
+++ b/cdp_backend/bin/process_local_file.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import argparse
+import logging
+import sys
+import traceback
+from pathlib import Path
+
+import fireo
+
+from cdp_backend.database import DATABASE_MODELS
+
+###############################################################################
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="[%(levelname)4s: %(module)s:%(lineno)4s %(asctime)s] %(message)s",
+)
+log = logging.getLogger(__name__)
+
+###############################################################################
+
+
+class Args(argparse.Namespace):
+    def __init__(self) -> None:
+        self.__parse()
+
+    def __parse(self) -> None:
+        p = argparse.ArgumentParser(
+            prog="clean_cdp_database",
+            description="Delete all collections in a CDP database.",
+        )
+        p.add_argument(
+            "google_credentials_file",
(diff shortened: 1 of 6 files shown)


cdp_backend/bin/process_local_file.py
```diff
@@ -0,0 +1,108 @@
...
+            # Create session from video_uri
+            session = Session(
+                # TODO: Should we instead take in datetime as an input?
+                session_datetime=datetime.utcnow(),
+                video_uri=video_uri,
+                session_index=0,
+            )
+
+            # Attach session to event
+            ingestion_model.sessions = [session]
```
JacksonMaxfield, line 86: Technically it's `session.video_uri`, a URI can refer to a local resource. Which would mean we can just accept the `EventIngestionModel` in full.

It also means we can process multi-session events. In which case I would change this to `process_event_from_local_files`?

Thoughts?

JacksonMaxfield: We would still need the upload but I guess what I am saying is:

1. Just accept the JSON
2. For each session, upload the local file and replace the `video_uri` with the uploaded files remote URI.
3. Send off the updated EventIngestionModel to the pipeline.

cdp_backend/utils/file_utils.py
```diff
@@ -81,6 +82,10 @@ def resource_copy(
     if dst is None:
         dst = uri.split("/")[-1]
 
+    # Create tmp directory to save file in
+    dirpath = tempfile.mkdtemp()
+    dst = Path(dirpath) / dst
```
JacksonMaxfield, line 87: I prefer:

```python
with tempfile.TemporaryDirectory() as temp_dir:
    ...
```

Because its a context manager it fully cleans up context close.

JacksonMaxfield (commented):
1. it's pretty neat that for this many lines of code we are adding a new feature, like go team on making it robust like this.
2. one comment on general design of how to accept this file and pass to pipeline
3. one nitpick comment on tempdir management, can ignore if you want.

Other than that, looks pretty sane to me.

JacksonMaxfield:
OH! And when attaching the URI back to the session model, add the `?alt=media` to the end. Unless that breaks the pipeline?

cdp_backend/bin/process_local_file.py
isaacna: Sounds good, will make that change

isaacna:
> OH! And when attaching the URI back to the session model, add the `?alt=media` to the end. Unless that breaks the pipeline?

Unfortunately this does break the pipeline when we try to do `resource_copy` on the file in GCS. Just curious, but what was the exact reason for appending that to the URI?

cdp_backend/utils/file_utils.py
isaacna: Also I realized if we use `with` and the put the rest of the function in the body, the file gets deleted before its used later in the pipeline and the pipeline will error out. The way I was using it before would delete the temp file, but the temp directory would still remain.

isaacna: One thing I can think of is to use create the temp directory in the event gather pipeline itself so the context doen't get deleted prematurely, but I think it'd make the code messier

JacksonMaxfield:
> > OH! And when attaching the URI back to the session model, add the `?alt=media` to the end. Unless that breaks the pipeline?
> 
> Unfortunately this does break the pipeline when we try to do `resource_copy` on the file in GCS, we get a `FileNotFoundError`. Just curious, but what exactly was the reason for appending that to the URI?

Oh whoops. I event added it in the wrong place myself. It is needed for google file storage to recognize that you aren't trying to download or copy the file but rather stream the file.

But what I meant by "I added it in the wrong place" is that you can keep it how it is for all the file storage and processing and such (exactly how it is now) but then during the database upload task, can you pass a parameter that says like "from_local" or something, and if it is local the session_video_uris need to be converted to their `https://` equivalents. `https://storage.googleapis.com....` Which is fortunately really easy to get:

```python
In [1]: from gcsfs import GCSFileSystem

In [2]: fs = GCSFileSystem(token="../cdp-backend/.keys/cdp-jackson-dev-002-sa-dev.json")

In [3]: fs.url("cdp-jackson-dev-002.appspot.com/9af4adb4f28b34c0142a2ab28fa5603dd5fdb347c074d957d13636dae554ed34-audio.wav")
Out[3]: 'https://storage.googleapis.com/download/storage/v1/b/cdp-jackson-dev-002.appspot.com/o/9af4adb4f28b34c0142a2ab28fa5603dd5fdb347c074d957d13636dae554ed34-audio.wav?alt=media'
```

cdp_backend/utils/file_utils.py
JacksonMaxfield: I am a bit confused... why would the tempdir be needed to prevent the original file from being removed? Don't we do a resource copy before every file operation?

JacksonMaxfield: Also the original file at this point has been uploaded to GCSFS no?

isaacna: I don't think I explained it well earlier, but if the user provides provides a video file to the script that lives in `cdp_backend`, then `resource_copy` fails with `FileExistsError`. 

For this edge case I guess I could just set `remove_local` to true when calling `upload_file` to avoid this but wasn't sure if that would be a good user experience. I thought using `tempdir` instead would solve all these issues but I'm having trouble getting the directory to cleanup after the whole workflow.

isaacna: Overall it's a pretty trivial error that doesn't matter if we're fine deleting the local copy with after uploading. Another really hacky thing we could also do is let the local copy get deleted and then call `resource_copy` again to fetch it, but I don't really like that idea lol

isaacna:
> > > OH! And when attaching the URI back to the session model, add the `?alt=media` to the end. Unless that breaks the pipeline?
> > 
> > 
> > Unfortunately this does break the pipeline when we try to do `resource_copy` on the file in GCS, we get a `FileNotFoundError`. Just curious, but what exactly was the reason for appending that to the URI?
> 
> Oh whoops. I event added it in the wrong place myself. It is needed for google file storage to recognize that you aren't trying to download or copy the file but rather stream the file.
> 
> But what I meant by "I added it in the wrong place" is that you can keep it how it is for all the file storage and processing and such (exactly how it is now) but then during the database upload task, can you pass a parameter that says like "from_local" or something, and if it is local the session_video_uris need to be converted to their `https://` equivalents. `https://storage.googleapis.com....` Which is fortunately really easy to get:
> 
> ```python
> In [1]: from gcsfs import GCSFileSystem
> 
> In [2]: fs = GCSFileSystem(token="../cdp-backend/.keys/cdp-jackson-dev-002-sa-dev.json")
> 
> In [3]: fs.url("cdp-jackson-dev-002.appspot.com/9af4adb4f28b34c0142a2ab28fa5603dd5fdb347c074d957d13636dae554ed34-audio.wav")
> Out[3]: 'https://storage.googleapis.com/download/storage/v1/b/cdp-jackson-dev-002.appspot.com/o/9af4adb4f28b34c0142a2ab28fa5603dd5fdb347c074d957d13636dae554ed34-audio.wav?alt=media'
> ```
> 
> It even adds that `?alt=media` to the URL!
> 
> TL;DR: it's need for the front-end streaming of the file and not the backend processing. So only change it right before database storage of the URI.

Ah gotcha that makes sense, I didn't know it had that use case for the front end. Will do

isaacna:
Just curious but does the front end have a way of streaming the video of something stored in GCS using the HTTP url?

isaacna:
Not sure why the `graphviz` fails to install on ubuntu. I added some changes to the github workflow, but I think the workflow on this PR runs off of the main `cdp_backend` and not my changes.

JacksonMaxfield:
> Just curious but does the front end have a way of streaming the video of something stored in GCS using the HTTP url?

Yep! See: 
Screen Shot 2021-08-19 at 07 47 11

From: https://councildataproject.org/seattle/#/events/2ee8dfd2-a4a5-416f-aded-f48b37e6e521

JacksonMaxfield:
> Not sure why the `graphviz` fails to install on ubuntu. I added some changes to the github workflow, but I think the workflow on this PR runs off of the main `cdp_backend` and not my changes.

They may have released something and something somewhere is broken in the dep chain, I will wait a bit and retrigger GitHub Actions to run.

cdp_backend/database/validators.py
```diff
@@ -115,7 +115,7 @@ def resource_exists(uri: Optional[str]) -> bool:
         return True
 
     # TODO Replace after finding way to pass custom fs through FireO validator
-    if uri.startswith("gs://"):
+    if uri.startswith("gs://") or uri.startswith("https://storage.googleapis"):
         return True
```
JacksonMaxfield, line 119: My comment about public read literally just solves this.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -102,10 +104,18 @@ def create_event_gather_flow(
...
         )
-        events: List[EventIngestionModel] = get_events_func(
-            from_dt=from_datetime,
-            to_dt=to_datetime,
-        )
+
+        # Use prefetched events instead of get_events_func if provided
+        # Or should we process in addition to function supplied events?
+        if prefetched_events is not None:
+            events = prefetched_events
```
JacksonMaxfield, line 111: Good question. I say no, keep it one or the other. The `get_events_func` takes a long time to run sometimes so this should be thought of as "I specifically want to process this event". Feel free to remove this comment question.

cdp_backend/bin/process_local_file.py
```diff
@@ -0,0 +1,94 @@
...
+
+            # Upload video file to file store
+            video_uri = upload_file(
+                credentials_file=config.google_credentials_file,
+                bucket=config.validated_gcs_bucket_name,
+                filepath=ingestion_model.sessions[0].video_uri,
+            )
+
+            # Replace video_uri of session
+            ingestion_model.sessions[0].video_uri = video_uri
```
JacksonMaxfield, line 72: Need to for loop this. There can be multiple sessions / multiple videos.

cdp_backend/utils/file_utils.py
JacksonMaxfield: You know what's annoying is that if we had public read setup on our stack all these issues around "File existence in GCS" would be solved. As soon as public read is solved, like three other issues are just gracefully handled.

(I have an open ticket with Google Cloud Support regarding a bug I found that makes it impossible to set public read on first attempt. Very interesting bug.)

Anyway, I think I understand now... but now this is outdated. Can you explain the path the file takes now that this was removed?

It gets uploaded and then downloaded each time? -- I can also try this branch locally today.

isaacna: Decided to add back in the tempdir stuff as per our offline conversation

cdp_backend/bin/process_local_file.py
isaacna: Oops yeah my bad, this was from when the script accepted the video file as an arg, but now we can actually support multiple!

JacksonMaxfield (approved):
Going to keep this open a second and test the ubuntu graphviz stuff tomorrow morning. But all looks good.

JacksonMaxfield:
This actually looks like a github actions release failure.... GitHub released a new build of their ubuntu runner 6 days ago which aligns with when the ubuntu tests started failing.... https://github.com/actions/virtual-environments/releases/tag/ubuntu20%2F20210816.1

JacksonMaxfield:
It's actually not `graphviz` its `ffmpeg` it looks like.... The `docs` test works just fine. Investigating.

JacksonMaxfield:
Got it figured out. Going to merge this then merge my PR to fix the builds without merge conflicts.

PR #97, opened 2021-09-01
feature/transcript-annotation-cleanup-and-section-annotation-type

JacksonMaxfield:
### Description of Changes

I got to spend some time on a couple of train rides starting to work on the section annotation problem. https://github.com/CouncilDataProject/cdp-roadmap/issues/9

Along the way, I figured I should clean up and better define our annotation problem. This is my v1 stab at it and it has worked so far. I generally like the API.

```python
transcript.annotations = TranscriptAnnotations(sections=[SectionAnnotation(), ]
```

For when there are no prior annotations and simple property setting when some annotations exist:

```python
transcript.annotations.sections = []
# OR
transcript.annotations.sections.append(SectionAnnotation())
```

### DIFF
--- a/cdp_backend/pipeline/transcript_model.py
+++ b/cdp_backend/pipeline/transcript_model.py
@@ -57,8 +57,8 @@ class SectionAnnotation:
     Notes
     -----
     The attributes `start_sentence_index` and `stop_sentence_index` should be treated
-    as inclusive and exclusive respectively, exactly like how Python `slice` function
-    works.
+    as inclusive and exclusive respectively, exactly like how the Python `slice`
+    function works.
 
     I.e. given a transcript with ordered sentences, the sentence indices can work
     as the parameters for a slice operation:
--- a/cdp_backend/pipeline/transcript_model.py
+++ b/cdp_backend/pipeline/transcript_model.py
@@ -139,9 +139,52 @@ class Transcript:
 # Docstrings for constants go _after_ the constant.
 
 
+@dataclass_json
+@dataclass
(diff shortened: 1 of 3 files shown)


JacksonMaxfield:
Oh! Notably, with this method where all objects are `dataclass` / `dataclass_json` all annotations can be fully serialized and deserialized! The implementation on `main` currently fails at deserialization back to typed objects because it uses `Enum` objects.

cdp_backend/pipeline/transcript_model.py
```diff
@@ -3,11 +3,105 @@
...
+    description: Optional[str] = None
+
+
+@dataclass_json
+@dataclass
+class TranscriptAnnotations:
+    """
+    Annotations that can appear (but are not guaranteed) for the whole transcript.
+    """
+
```
isaacna, line 101: Should we add paramter docstrings for `sections`?

isaacna (approved):
Looks good, I think the model makes sense! I personally like the JSON serializable classes much more than dictionaries.

cdp_backend/pipeline/transcript_model.py
JacksonMaxfield: I actually removed it because I ran `make docs` and saw that the typing creates a link to the `SectionAnnotation` object documentation. Like just having this typing links to the other docs so I ignored.

JacksonMaxfield: Going to merge now but can change in a later PR if we want more docs.

PR #101, opened 2021-09-07
feature/resources-exists-from-gcs-storage

### LINKED ISSUE #52: Add back `resource_exists` validator for `uri` field on File
We had to remove the usage of `resources_exists` from the uri field validator because we couldn't think of a method to pass the google credentials to the validator. At some point we may want to add this back but this shouldn't hold up #49 from being merged.

> Actually after trying this, I think passing another param to the validator is a little tricky. 

> I can add kwargs to `resource_exists`, but I'm not sure if there's an easy way to pass in `fs` into that function inside `uri = fields.TextField(required=True, validator=validators.resource_exists)`. 

> Because model fields in FireO are assigned like `db_file.uri = "gs://stuff"`, I'm not sure we can pass in other params into the `validators` arg. 

> Would it make more sense to just validate separately outside of FireO's built-in field validation? Or is there an easier way to do this I'm missing?

_Originally posted by @isaacna in https://github.com/CouncilDataProject/cdp-backend/pull/49#discussion_r589185839_

isaacna:
### Link to Relevant Issue

This pull request resolves #52 

### Description of Changes
- Allowing validating GCS uri's by passing in the credentials file since FireO now supports `validator_kwargs`
- Created a method to transform the GCS JSON API format to a gsutil formatted string, because that's what `GCSFileSystem` works with
- Added a dependency on FireO by installing via github, with the commit specifically for the changes when I [added the kwargs](https://github.com/isaacna/FireO/commit/d6b47f8d5c62025f786559ff68b327521acb14f2)
  - We have to do this the 1.4.5 version which contains my changes hasn't been released to PyPi yet. I already have an [issue](https://github.com/octabytes/FireO/issues/125) open for that and will change it to 1.4.5 once that happens 

### Testing
- Ran a run of the local file processing script which uses GCS uri's and verified that they are being validated properly

### DIFF
--- a/cdp_backend/tests/utils/test_string_utils.py
+++ b/cdp_backend/tests/utils/test_string_utils.py
@@ -27,3 +27,13 @@ from cdp_backend.utils import string_utils
 )
 def test_clean_text(text: str, expected: str, clean_stop_words: bool) -> None:
     assert string_utils.clean_text(text, clean_stop_words=clean_stop_words) == expected
+
+@pytest.mark.parametrize(
+    "text, expected",
+    [
+        ("https://storage.googleapis.com/download/storage/v1/b/bucket.appspot.com/o/wombo_combo.mp4?alt=media", "gs://bucket.appspot.com/wombo_combo.mp4"),
+        ("https://storage.googleapis.com/download/storage/bucket.appspot.com/o/wombo_combo.mp4?alt=media", ""),
+    ],
+)
+def test_convert_gcs_json_url_to_gsutil_form(text: str, expected: str) -> None:
+    assert string_utils.convert_gcs_json_url_to_gsutil_form(text) == expected
--- a/cdp_backend/tests/utils/test_string_utils.py
+++ b/cdp_backend/tests/utils/test_string_utils.py
@@ -28,11 +28,21 @@ from cdp_backend.utils import string_utils
 def test_clean_text(text: str, expected: str, clean_stop_words: bool) -> None:
     assert string_utils.clean_text(text, clean_stop_words=clean_stop_words) == expected
 
+
 @pytest.mark.parametrize(
     "text, expected",
     [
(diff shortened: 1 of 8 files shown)


cdp_backend/tests/utils/test_string_utils.py
```diff
@@ -27,3 +27,23 @@
...
     assert string_utils.clean_text(text, clean_stop_words=clean_stop_words) == expected
+
+
+@pytest.mark.parametrize(
+    "text, expected",
+    [
+        (
+            "https://storage.googleapis.com/download/storage/v1/b/"
+            + "bucket.appspot.com/o/wombo_combo.mp4?alt=media",
+            "gs://bucket.appspot.com/wombo_combo.mp4",
```
JacksonMaxfield, line 38: WHERE YOU AT?? Classic.

JacksonMaxfield (commented):
This all looks good. My only major comment is that the only place this should be needed to my knowledge is on the `File` model. All other URIs in our DB link to resources outside of our control.

Can reduce the changeset if you want, your call.

Will wait for fireo release to merge though probably.

cdp_backend/tests/utils/test_string_utils.py
isaacna: haha glad you

> This all looks good. My only major comment is that the only place this should be needed to my knowledge is on the `File` model. All other URIs in our DB link to resources outside of our control.
> 
> Can reduce the changeset if you want, your call.
> 
> Will wait for fireo release to merge though probably.

I'm more inclined to keep the changes on the models just in case that changes, even if it's unlikely. And yeah, I'm fine with keeping the PR open until FireO 1.4.5 is actually released

JacksonMaxfield (approved):
Looks good to me! Thanks for working with FireO. We will likely request more changes in the future / features so nice to know we have a good working relationship 👍

JacksonMaxfield:
Merge whenever tests pass @isaacna

PR #105, opened 2021-09-09
feature/process-special-event

isaacna:
### Link to Relevant Issue

- Relevant `cdp-backend` changes for this [cookiecutter feature request](https://github.com/CouncilDataProject/cookiecutter-cdp-deployment/issues/46)

### Description of Changes

- Changing `process_local_file` to `process_special_event`
- If the `video_uri` is a remote resource, we'll copy that file locally and then upload it to the file store 
- Lmk if we want a more robust way of checking if it's remote resource. I just did a simple `http://` string check but am open to suggestions for improvement

### DIFF
--- a/cdp_backend/bin/process_special_event.py
+++ b/cdp_backend/bin/process_special_event.py
@@ -11,6 +11,7 @@ from cdp_backend.file_store.functions import upload_file
 from cdp_backend.pipeline import event_gather_pipeline as pipeline
 from cdp_backend.pipeline.ingestion_models import EventIngestionModel
 from cdp_backend.pipeline.pipeline_config import EventGatherPipelineConfig
+from cdp_backend.utils.file_utils import resource_copy
 
 ###############################################################################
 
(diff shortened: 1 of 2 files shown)


cdp_backend/bin/process_special_event.py
```diff
@@ -62,11 +63,18 @@ def main() -> None:
             )
 
             for session in ingestion_model.sessions:
+                # Copy if remote resource, otherwise use local file uri
+                if session.video_uri.startswith("http://") or session.video_uri.startswith("https://"):
+                    filepath = resource_copy(session.video_uri)
+
+                else:
+                    filepath = session.video_uri
```
JacksonMaxfield, line 71: The safest way to do this would be to do:

```python
from fsspec.core import url_to_fs
from fsspec.implementations.local import LocalFileSystem

fs, path = url_to_fs(filepath)
if not isinstance(fs, LocalFileSystem):
    # copy remote
```

JacksonMaxfield (approved):
Looks good to me. I would personally use `ffspec`'s file system determiner function instead of the strings on the path. 🤷

PR #110, opened 2021-09-20
Only use temp directory if dst not defined

isaacna:
### Description of Changes

- Fixing a bug where we were creating a temporary directory everytime in `resource_copy` [here](https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/utils/file_utils.py#L84-L89) instead of directly using `dst` when provided
- Now we only create the temp directory if `dst` is not defined to avoid resource copying directly into `cdp_backend`. [Context](https://github.com/CouncilDataProject/cdp-backend/pull/89#discussion_r691814155) for why this matters

### DIFF
--- a/cdp_backend/utils/file_utils.py
+++ b/cdp_backend/utils/file_utils.py
@@ -84,9 +84,9 @@ def resource_copy(
     if dst is None:
         dst = uri.split("/")[-1]
 
-    # Create tmp directory to save file in
-    dirpath = tempfile.mkdtemp()
-    dst = Path(dirpath) / dst
+        # Create tmp directory to save file in
+        dirpath = tempfile.mkdtemp()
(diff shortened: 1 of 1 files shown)


JacksonMaxfield (commented):
I like having `None` default to "current directory". Whatever was causing the original bug, could you simply create a tempdir and pass a dest to this function?

I like having "None" mean current dir

isaacna:
> I like having `None` default to "current directory". Whatever was causing the original bug, could you simply create a tempdir and pass a dest to this function?

Sounds good, I'm fine with doing that since I think it'll make `resource_copy` more true to it's original intent and documentation

JacksonMaxfield (approved):
Perfect!

PR #112, opened 2021-09-27
feature/simplified-video-download-and-processing

### LINKED ISSUE #102: Combine all the video processing tasks into a single task
### Feature Description

Reduce the pipeline computation time by combining all the tasks that require the video on the worker together.

### Use Case

Drastically reducing the pipeline computation time.

### Solution

Move the function calls to generate thumbnails to the start of the pipeline with the strip audio and generate hash.

### Alternatives

Shak2000:
### Link to Relevant Issue

This pull request resolves #102 

### Description of Changes

This pull request would streamline the procedure for copying, processing, and deleting video files used to generate thumbnails and audio transcripts.

### DIFF
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -126,9 +126,7 @@ def create_event_gather_flow(
             session_processing_results: List[SessionProcessingResult] = []
             for session in event.sessions:
                 # Get or create audio
-                tmp_video_filepath = resource_copy_task(
-                    uri=session.video_uri
-                )
+                tmp_video_filepath = resource_copy_task(uri=session.video_uri)
 
                 session_content_hash, audio_uri = get_video_and_split_audio(
                     tmp_video_filepath=tmp_video_filepath,
@@ -177,7 +175,10 @@ def create_event_gather_flow(
                     credentials_file=config.google_credentials_file,
                 )
 
-#                fs_functions.remove_local_file(tmp_video_filepath)
+                # Add audio uri and static thumbnail uri
+                resource_delete_task(
+                    tmp_video_filepath, audio_uri, static_thumbnail_uri
+                )
 
                 # Store all processed and provided data
                 session_processing_results.append(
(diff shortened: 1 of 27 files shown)


cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -715,7 +742,7 @@ def generate_transcript(
 @task(nout=2)
 def get_video_and_generate_thumbnails(
     session_content_hash: str,
-    video_uri: str,
+    tmp_video_path: str,
```
isaacna, line 745: Let's update the docstrings to match this

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -197,17 +204,50 @@ def create_event_gather_flow(
...
+        uri=uri,
+        dst=dirpath,
+        overwrite=True,
+    )
+
+    return str(Path(dirpath) / uri.split("/")[-1])
+
+
+@task
+def resource_delete_task(uri: str, audio_uri: str, static_thumbnail_uri: str) -> None:
```
isaacna, line 222: Instead of including `audio_uri` and `static_thumbnail_uri` as params to the function, you could use [`upstream_tasks`](https://docs.prefect.io/core/concepts/execution.html#triggers) instead while calling `resource_delete_task`. This way we can make sure the flow happens in order without adding params to the function signature that aren't really relevant to the function logic itself.

So I think it could be something like:
```
audio_uri = get_video_and_split_audio(...)
static_thumbnail_uri = get_video_and_generate_thumbnails(...)

resource_delete_task(upstream_tasks=[audio_uri, static_thumbnail_uri])
```

isaacna (commented):
Looks good to me, mostly just some minor prefect stuff. But I was slightly unclear about what the original ask of #102 is. Is removing the redundant `resource_copy` call the only thing you wanted to fix @JacksonMaxfield ?

JacksonMaxfield (commented):
Agree with all of @isaacna's comments.

Update docstrings, and swap parameters for `upstream_tasks` Prefect param.

If I was to ask for one more thing I would say that we should just split up the `get_video_and_split_audio` task into two functions / tasks. One that generates the session content hash and one that splits the audio.

---

As a note, my original intent of the GitHub issue was to literally just shove the thumbnail functions into the `get_video_and_split_audio` task but that seems clunky.... my reasoning for doing that goes to the "well if we ever wanted to run the pipeline with many machines then each task needs to do all of it's own IO" but we haven't ran the pipeline on anything but a single machine so far so this is totally fine. Let's do the simple thing now and add complexity only when needed.

isaacna (approved):
Looks good!

JacksonMaxfield (approved):

PR #114, opened 2021-09-28
admin/simpler-ci-and-build-infra

JacksonMaxfield:
### Description of Changes

Simplifies the CI / build infrastructure. Everything is now routed through `tox` so GitHub actions now just call tox instead of duplicating commands. Also switches to pre-commit for linting. This should make the process of contribution easier because someone can run `tox -e lint` and have it attempt to fix most mistakes for them, then run it again and ensure that mistakes _were_ fixed.

### DIFF
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -48,9 +48,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install .[test]
-    - name: Test with pytest
-      run: |
-        pytest --cov-report xml --cov=cdp_backend cdp_backend/tests/
+    - name: Run tests with Tox
+      run: tox -e py
     - name: Upload Codecov
       uses: codecov/codecov-action@v1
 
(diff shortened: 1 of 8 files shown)


.github/workflows/build-main.yml
```diff
@@ -70,19 +69,9 @@ jobs:
...
-        isort cdp_backend --check-only
-    - name: Check typing with mypy
-      run: |
-        mypy cdp_backend
-    - name: Check with black
-      run: |
-        black --check cdp_backend
+        pip install pre-commit
+    - name: Lint
+      run: pre-commit run --all-files --show-diff-on-failure
```
isaacna, line 74: Just checking, but does this run on everything in `cdp_backend` or only `cdp_backend/cdp_backend`?

isaacna (approved):
Looks good! Just had one question about how what directory level pre-commit runs on

.github/workflows/build-main.yml
JacksonMaxfield: Good question. I think it runs on the whole repo but thats personally fine by me.

isaacna: Gotcha, my only comment would be sometimes I store things in the `cpd_backend` directory like files I'm going to run the pipeline on or various config files which could get in the way. Or maybe is there a way to only check files/changes that are added to git?

JacksonMaxfield: It should only be running lint and stuff on python files. Configs and such are ignored by black / flake8

JacksonMaxfield: Going to merge but if you encounter issues with this behaviour in the future let me know and I will figure out how to specify this to just the `cdp-backend/cdp_backend` dir

PR #115, opened 2021-09-29
feature/turn-off-db-model-autoload-and-update-hasher

### LINKED ISSUE #106: Disable ReferenceField `auto_load` for all Models
### Feature Description

Currently `ReferenceField` properties for all of our models are loading on parent model initialization. What this means is when I request a `Vote` from the database, it will first load the vote data, then for each reference field load the referenced model, then recursively do that down the model dependency tree. This is great for "simple" access but terrible for large scale access.

### Use Case

I want to be able to query the database for all votes and not have my calls rejected by Firebase.

### Solution

Turn off `auto_load`: https://octabyte.io/FireO/fields/reference-field/#auto-load

Additionally, because certain parts of our pipeline depend on / expect those referenced models to exist, the pipeline will likely need to be updated to fetch the referenced model when needed.

### Alternatives

isaacna:
### Link to Relevant Issue

This pull request resolves #106 

### Description of Changes

Turned off `auto_load` for all reference fields in the db models. The only part that gets affected should be when we recursively generate the hash for a document. 

When the recursion starts going more than 1 level deep (example: `Session` -> `Event` -> `Body`), `Body` will be typed as a `ReferenceDocumentLoader` rather than a `Model` when `auto_load` is off. That's why I added the extra if statement to hash based on the `id` of the fetched reference document in that case.

### DIFF
--- a/cdp_backend/database/functions.py
+++ b/cdp_backend/database/functions.py
@@ -9,6 +9,7 @@ from typing import Any, List, Optional
 
 import fireo
 from fireo.models import Model
+from fireo.queries.query_wrapper import ReferenceDocLoader
 from google.cloud.firestore_v1.batch import WriteBatch
 from google.cloud.firestore_v1.transaction import Transaction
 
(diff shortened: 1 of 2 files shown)


JacksonMaxfield (approved):
Dang. This is perfect!

PR #118, opened 2021-10-05
feature/add-indexes-for-role-models

JacksonMaxfield:
### Description of Changes

Was working with Brian to make sure certain parts of the database are queryable for frontend stuff. These aren't _needed_ but are nice to have.

### DIFF
--- a/cdp_backend/database/models.py
+++ b/cdp_backend/database/models.py
@@ -209,7 +209,60 @@ class Role(Model):
         return role
 
     _PRIMARY_KEYS = ("title", "person_ref", "body_ref", "seat_ref")
-    _INDEXES = ()
+    _INDEXES = (
+        IndexedFieldSet(
+            (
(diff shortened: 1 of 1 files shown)


isaacna (approved):
Looks good to me! 

Also, I was thinking if it'd be worth creating some helper method that given some `IndexField`'s, spits out a set of `IndexedFieldSet` with all or multiple permutations of `order` to reduce code repetition?

isaacna:
> Also, I was thinking if it'd be worth creating some helper method that given some `IndexField`'s, spits out a set of `IndexedFieldSet` with all or multiple permutations of `order` to reduce code repetition?

Actually nvm I don't like this idea. I think having the index sets explicitly defined is probably better for a model class, and the exact combinations of order might be specific to the front end so creating them programmatically probably isn't worth the hassle

Issue #119, opened 2021-10-06
Debug why some files can't be archived / linked


evamaxfield:
### Describe the Bug

_A clear and concise description of the bug._

Was checking the logs of test-deployment and saw that some files fail to be archived: https://github.com/CouncilDataProject/test-deployment/runs/3776409036?check_suite_focus=true#step:8:118

I can click on that link and it is a real file that is publically accessibile.

### Expected Behavior

_What did you expect to happen instead?_

All publically accessible files are archivable.

JacksonMaxfield:
@isaacna I think this may also be timeout errors or similar: https://github.com/intake/filesystem_spec/issues/619#issuecomment-825029746

I am seeing a lot of failures on seattle-staging backfill runs because of timeout errors: https://github.com/CouncilDataProject/seattle-staging/runs/3875382085?check_suite_focus=true#step:9:157

isaacna:
Aren' the timeout errors what'

> It honestly may be just that we are storing that file and then _instantly_ checking if it exists. I think most storage systems are "eventually consistent" / it may take some time for Google to properly hit the result after storing. (Especially when we are uploading hundreds of files at a time...)

I'm kinda confused about this, but for this specific case isn't it checking that the remote resource `https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf` exists via an http request? Or am I misunderstanding what `fsspec.url_to_fs` does? 

I was thinking that sometimes there's just an error while making the http request (for the non google uri's): https://github.com/intake/filesystem_spec/blob/master/fsspec/implementations/http.py#L305-L315

Also for the timeout errors during `resource_copy` I dont' think that's related to the validator issue but still a problem. Maybe we could add some retry logic into `resource_copy`?

isaacna:
I also created a script that just checks creates a file store using `fsspec`, and checks whether the resource exists. I ran it in a loop but haven't seen a case with the resource suddenly not being found. Maybe the github action runner is just more prone to http errors or timeouts?

isaacna:
It looks like Event gather runs 160-164 (the ones since bumping the cdp-backend version) haven't encountered the `could not be archived` issue. I'll wait for a few more runs before saying that the HEAD request fixed the issue, but fingers crossed that it did

isaacna:
[...]

PR #121, opened 2021-10-10
debug/error-logging-for-field-validation-failed

isaacna:
### Link to Relevant Issue

Part of #119

### Description of Changes

I wasn't able to recreate the failed archiving example listed in the issue, so I decided to add some logging statements to help debug (and I think it'll be good to have in general). I haven't seen the error occur when I ran the pipeline using the mock data manually, so this should help if it happens in the one of the runs again.

There are a [few different cases](https://github.com/octabytes/FireO/blob/master/src/fireo/fields/field_attribute.py#L99-L115) where FieldValidationFailed is thrown, so the log statement will help narrow down what's causing the validation to fail when we don't expect it to.


JacksonMaxfield:
Or actually, you should just feel free to merge after lint is fixed.

PR #122, opened 2021-10-13
feature/use-head-request-in-resource-exists

isaacna:
### Link to Relevant Issue

Related to #119 

### Description of Changes

Using HEAD request via `requests` library instead of `fsspec` since their implementation uses a GET request as described in the parent issue. Hopefully this will help resolve the transient issues we run into with `resource_exists` since we the operation is less expensive.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,4 +25,4 @@ repos:
           - id: mypy
             additional_dependencies: 
               - "types-pytz"
-              - "requests==2.26.0"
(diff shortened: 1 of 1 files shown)


isaacna:
Not sure why `mypy` is still failing. The `Library stubs not installed for` seems to be a known issue in mypy precommit, which should be resolved by adding it to `additional_dependencies` (--install-types doesn't work in pre commit) according to this:
https://jaredkhan.com/blog/mypy-pre-commit

PR #124, opened 2021-10-20
feature/add-more-logging-to-special-event-processing

evamaxfield:
### Description of Changes

When processing our CDP event last night I was confused as to if anything was happening or if the process was hung. This adds a tiny bit more logging to the process for user feedback.


JacksonMaxfield:
This is just to clarify pipeline behavior. No bugs!

PR #127, opened 2021-10-27
bugfix/store-files-to-fs-with-content-hash-as-name

### LINKED ISSUE #120: Store files to unique paths instead of filenames
### Feature Description

For certain files that we want to archive (such as person picture) we are simply downloading and then uploading to our own file storage and storing at `/[filename]`. Instead let's store the file at a unique path so that we can maybe reduce the amount of uploads.

### Use Case

In addition to reducing the number of uploads we run, this should also just speed up the pipeline, and crucially: make our data storage "safer", in the case that a person uses the same filenaming for all of their pictures i.e. `council.gov/{PERSONNAME}/picture.png` the current pipeline would result in only storing a single picture for them all instead of storing them uniquely.

### Solution

Anytime we are using the `download` -> `upload` pattern: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/event_gather_pipeline.py#L874

change it to `download` -> `hash` -> `check if exists (by hash)` -> `(optional) upload (with hash as the file name)`.

### Alternatives

Shak2000:
### Link to Relevant Issue

This pull request resolves #120

### Description of Changes

The program would now be able to add a hash at the beginning of the name of any file to be stored.
The program would now be able to delete temporary files after they are uploaded to the cloud.

### DIFF
--- a/cdp_backend/utils/file_utils.py
+++ b/cdp_backend/utils/file_utils.py
@@ -63,7 +63,6 @@ def resource_copy(
     uri: str,
     dst: Optional[Union[str, Path]] = None,
     overwrite: bool = False,
-    add_hash: bool = False,
 ) -> str:
     """
     Copy a resource (local or remote) to a local destination on the machine.
@@ -78,9 +77,6 @@ def resource_copy(
     overwrite: bool
         Boolean value indicating whether or not to overwrite a local resource with
         the same name if it already exists.
-    add_hash: bool
-        Boolean value indicating if the file needs a unique hash to be added to the
-        beginning of its name.
 
     Returns
     -------
@@ -89,9 +85,6 @@ def resource_copy(
     """
     if dst is None:
         dst = uri.split("/")[-1]
-        if add_hash:
-            hash = hash_file_contents(uri)
-            dst = hash + "-" + uri.split("/")[-1]
 
     # Ensure dst doesn't exist
     dst = Path(dst).resolve()
@@ -368,3 +361,22 @@ def hash_file_contents(uri: str, buffer_size: int = 2 ** 16) -> str:
             hasher.update(block)
 
     return hasher.hexdigest()
+
+
+def generate_file_storage_name(local_path: str) -> str:
+    """
+    Return the name of the file as it should be on Google Cloud Storage.
+
+    Parameters
+    ----------
+    local_path: str
+        The name of the file path on the local computer.
+
+    Returns
+    -------
+    dst: str
+        The name of the file as it should be on Google Cloud Storage.
+    """
+    hash = hash_file_contents(local_path)
+    dst = hash + "-" + local_path.split("/")[-1]
+    return dst
--- a/cdp_backend/utils/file_utils.py
+++ b/cdp_backend/utils/file_utils.py
@@ -363,20 +363,20 @@ def hash_file_contents(uri: str, buffer_size: int = 2 ** 16) -> str:
     return hasher.hexdigest()
 
 
-def generate_file_storage_name(local_path: str) -> str:
+def generate_file_storage_name(extension: str) -> str:
     """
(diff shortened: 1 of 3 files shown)


cdp_backend/utils/file_utils.py
```diff
@@ -83,6 +90,9 @@ def resource_copy(
     """
     if dst is None:
         dst = uri.split("/")[-1]
+        if addHash:
+            hash = str(uuid.uuid4())
+            dst = hash + "-" + uri.split("/")[-1]
```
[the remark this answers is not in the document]
JacksonMaxfield: Clarifying: This may be better solved on the upload side rather than the download side. When we upload, change the filename to just the hash or something similar since the download seems to be working just fine.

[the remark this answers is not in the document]
Shak2000: Do you have an example of a configuration that would result in the naming problem described above?

I can definitely move the hash generation code to the upload side, but it would not be standard compared to the rest of the system. WDYT?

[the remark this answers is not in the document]
JacksonMaxfield: https://github.com/CouncilDataProject/cdp-backend/issues/120#issuecomment-944473376

The link above is pulled directly from a pipeline log. I forget which one but we have encountered that error before.

I would say don't change the upload function so much as add a new function to generate an "upload name" that generates a hash and attaches it to some suffix.

in pseudocode in my head its something like:

```python
local_path = resource_copy(...)
upload_filename = generate_file_storage_name(local_path)
uploaded_uri = upload_file(..., save_name=upload_filename)
```

cdp_backend/utils/file_utils.py
```diff
@@ -359,3 +361,22 @@ def hash_file_contents(uri: str, buffer_size: int = 2 ** 16) -> str:
...
+    Parameters
+    ----------
+    local_path: str
+        The name of the file path on the local computer.
+
+    Returns
+    -------
+    dst: str
+        The name of the file as it should be on Google Cloud Storage.
+    """
```
JacksonMaxfield, line 379: my bad in pseduocode, we can't do the `local_path.split("/")` thing because this would still result in weird filenames.

if the localpath is `parent_dir/child_dir/original_uri?param1&param2` this function would result in:

`{hash}-original_uri?param1&param2`. Can you add a second parameter here that is just "suffix" so the end result is simply: `f"{hash}-{suffix}"`?

Shak2000: Since the path comes from the file name in the local file system, and the path is generated after `resource_copy` is called, the program does not include `param1&param2` in the value for the `local_path` parameter. For example, I get `/Users/shak/PycharmProjects/cdp-backend/Mosqueda_225x225.jpg` as a value for `local_path`.

JacksonMaxfield: Looking at the call stack:

`resource_copy` is called with just the origin URI, no dest, so it generates a dest by doing the `"/"` split. In which case the dest will be the current working directory and whatever is at the tail end of the URI. So the final local path after copy will be something with odd naming. Basically the local path is also just odd because the original URI includes params.

The test you ran with Mosqueda is correct but it doesn't have odd naming.

Shak2000: Done.

JacksonMaxfield: Wait wait wait. Sorry this is all misscommunication on my end. Very sorry!

I meant keep the hash by the file content, but add a suffix parameter.

So we can do `generate_file_storage_name(local_path: str, suffix: str) -> str:`

I think the files in google cloud should be named `{hash}-person-picture` and to generate that you would need to call `generate_file_storage_name(file, "person-picture")`.

What do you think?

Shak2000: I think that I fixed this issue based on what you proposed.

PR #130, opened 2021-11-04
bugfix/update-index-pipeline-and-search-to-non-recursive-db-models

JacksonMaxfield:
### Description of Changes

Somehow we totally missed some scripts and workflows when we merged #115, this has caused the index pipeline to completely fail for the last couple of days. This PR fixes all references to `_ref.` which is generally save because anywhere that is used is where we were previously doing `_ref.{property}` and we can now change it to `_ref.get()` or do that operation once and store it somewhere for reuse.

Also all of the workflow bin scripts now properly raise an error if the workflow state results in failed.

I will fix up the tests tomorrow.


cdp_backend/bin/search_cdp_events.py
```diff
@@ -57,6 +60,11 @@ def __parse(self) -> None:
             prog="search_cdp_events", description="Search CDP events given a query."
         )
 
+        p.add_argument(
+            "instance",
+            type=str,
+            help="Which instance to query.",
+        )
```
isaacna, line 67: Is this related to starting cross municipality API?

cdp_backend/bin/search_cdp_events.py
```diff
@@ -148,20 +156,19 @@ def run_remote_search(query: str, sort_by: str, first: int = 4) -> None:
...
 
     compiled_events: List[EventMatch] = []
-    for matching_grams in matching_events.values():
+    for event_id, matching_grams in matching_events.items():
         compiled_events.append(
             EventMatch(
-                event_ref=matching_grams[0].event_ref,
+                event=db_models.Event.collection.get(
+                    f"{db_models.Event.collection_name}/{event_id}"
+                ),
```
isaacna, line 171: Just to make sure, but do we iterate through items because `matching_grams` has an autoloaded reference to the event? I guess I wasn't clear on what the datatype of `matching_grams` is

isaacna (approved):
Thanks for catching this. I didn't test the event index pipeline or `search_cdp_events` as I didn't consider that turning off autoloading would affect them too. And adding the fail checks in the scripts is smart to prevent issues like this from happening silently in the future.

cdp_backend/bin/search_cdp_events.py
JacksonMaxfield: Huh. I added it to easily test the changes but I guess both :shrug:

cdp_backend/bin/search_cdp_events.py
JacksonMaxfield: `matching_grams` is `Dict[str, db_models.IndexedEventGram]`. Yep gotta get the matching event but on further look I think I can just call `indexed_event_gram.event_ref.get()`

cdp_backend/tests/pipeline/test_event_index_pipeline.py
```diff
@@ -50,66 +52,77 @@ def test_create_event_index_flow(n_grams: int, store_local: bool) -> None:
 
 #############################################################################
 
+
+def apply_mocking_funcs_and_refs(model: Model) -> Model:
+    model = db_functions.generate_and_attach_doc_hash_as_id(model)
+    model.ref = firestore.DocumentReference(model.collection_name, model.id)
+    model.get = lambda: model
+    return model
```
JacksonMaxfield, line 60: @isaacna calling your attention to this helpful function I wrote to setup any db model with mocks / correct data

PR #134, opened 2021-11-10
feature/optimize-pipeline

JacksonMaxfield:
### Description of Changes

This PR includes both bugfixes and event gather pipeline optimizations for computation time.

1. The `_process_person` function now caches prior persons that have already been processed instead of always reprocessing them. This was the largest slowdown in the pipeline because downloading, uploading, and creating the DB models (and the person and seat pictures) took a really like time. (Note, these are cached on the per event level, not the whole pipeline level, so there is _some_ extra uploading but compared to before this is so minor)
2. Fixed a race condition that is affecting all instances index pipelines: https://github.com/CouncilDataProject/seattle-staging/runs/4160813568?check_suite_focus=true
3. Drastically changed the hover thumbnail generator function to instead of using a set step size to simply random pull frames from the video at a 1% of pulling each frame as it iterates through the video. This reduced the time it took to generate a hover thumbnail from multiple minutes at a time to seconds.

Overall I ran the pipeline for Seattle Staging before and after these changes:

Before: processing five events ~2 hours
After: processing five events ~25 minutes

### DIFF
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -915,33 +915,39 @@ def _process_person_ingestion(
     default_session: Session,
     credentials_file: str,
     bucket: str,
+    file_upload_cache: Dict[str, db_models.File] = {},
 ) -> db_models.Person:
     # Store person picture file
     person_picture_db_model: Optional[db_models.File]
+    person_picture_key = f"{person.name} -- person-picture"
     if person.picture_uri is not None:
         try:
-            tmp_person_picture_path = file_utils.resource_copy(
-                person.picture_uri,
-            )
-            destination_path = file_utils.generate_file_storage_name(
-                tmp_person_picture_path,
-                "person-picture",
-            )
-            person_picture_uri = fs_functions.upload_file(
-                credentials_file=credentials_file,
-                bucket=bucket,
-                filepath=tmp_person_picture_path,
-                save_name=destination_path,
-                remove_local=True,
-            )
-            person_picture_db_model = db_functions.create_file(
-                uri=person_picture_uri,
-                credentials_file=credentials_file,
-            )
-            person_picture_db_model = db_functions.upload_db_model(
-                db_model=person_picture_db_model,
-                credentials_file=credentials_file,
-            )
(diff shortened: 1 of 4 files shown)


cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -915,158 +915,177 @@ def _process_person_ingestion(
...
-                    if person_role.body.start_datetime is None:
-                        person_role_body_start_datetime = (
-                            default_session.session_datetime
-                        )
-                    else:
-                        person_role_body_start_datetime = (
-                            person_role.body.start_datetime
-                        )
+        # Update upload cache with person
+        upload_cache[person_cache_key] = person_db_model
```
isaacna, line 990: Isn't the type of `person_db_model` here `Person`? Shouldn't the type of `upload_cache` be `Dict[str, Model]` then?

cdp_backend/pipeline/event_index_pipeline.py
```diff
@@ -458,6 +458,23 @@ def create_event_index_pipeline(
         The constructed CDP Event Index Pipeline as a Prefect Flow.
     """
     with Flow("CDP Event Index Pipeline") as flow:
+        # Ensure stopwords are downloaded
+        # Do this once to ensure that we don't enter a race condition
+        # with multiple workers trying to download / read overtop one another
+        # later on.
+        try:
```
isaacna, line 465: Good catch! Is there a way to check if we have the stopwords downloaded without having to do a try catch?

isaacna (commented):
Sorry I couldn't get to this earlier. I know this is already merged, but just had one question about the parameter typing for the cache. Otherwise looks great and glad the event gather pipeline is running much faster now!

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -915,158 +915,177 @@ def _process_person_ingestion(
     default_session: Session,
     credentials_file: str,
     bucket: str,
+    upload_cache: Dict[str, db_models.File] = {},
```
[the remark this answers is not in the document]
JacksonMaxfield: Completely correct! Will push through a patch to change this

PR #135, opened 2021-11-13
feature/add-metadata-doc-to-db

### LINKED ISSUE #117: Add basic instance metadata collection to infrastructure stack
### Feature Description

Partially discussed in https://github.com/CouncilDataProject/cdp-roadmap/issues/3, it would be nice to have a collection with a single document to pull metadata about the instance.

### Use Case

Most importantly would be which cdp version the instance is currently on. This would help in determine which features the instance supports.

### Solution

Use pulumi gcp firestore document resource to create / update the same document every time the infrastructure is upgraded.
https://www.pulumi.com/docs/reference/pkg/gcp/firestore/document/

@tohuynh @isaacna can you think of other metadata that would be good to store in the database itself about the instance?
version, some subset of cookiecutter params maybe? anything else?

JacksonMaxfield:
### Link to Relevant Issue

This pull request resolves #117 

### Description of Changes

Adds basic metadata collection and a configuration doc in said collection.
image

### DIFF
--- a/cdp_backend/infrastructure/cdp_stack.py
+++ b/cdp_backend/infrastructure/cdp_stack.py
@@ -1,20 +1,34 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+import json
+
 import pulumi
 import pulumi_gcp as gcp
 from pulumi_google_native.firestore import v1 as firestore
 
 from ..database import DATABASE_MODELS
+from ..version import __version__
 
 ###############################################################################
+
+
+class GoverningBody:
+    city_council = "city council"
+    county_council = "county council"
+    school_board = "school board"
+    other = "other"
 
 
 class CDPStack(pulumi.ComponentResource):
     def __init__(
         self,
         gcp_project_id: str,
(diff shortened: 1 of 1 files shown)


cdp_backend/infrastructure/cdp_stack.py
```diff
@@ -1,20 +1,34 @@
...
+    county_council = "county council"
+    school_board = "school board"
+    other = "other"
+
+
 class CDPStack(pulumi.ComponentResource):
     def __init__(
         self,
         gcp_project_id: str,
+        municipality_name: str = "dev",
```
isaacna, line 27: should we make the default something like `"dev_municipality"`? Very minor but just to distinguish the default values between fields

isaacna (approved):
Looks good to me!

tohuynh (approved):
Looks good to me.

cdp_backend/infrastructure/cdp_stack.py
JacksonMaxfield: Done!

PR #137, opened 2021-11-24
admin/use-setup-python-caching

isaacna:
### Description of Changes

- Upgrading `setup-python` and `checkout` actions and adding pip caching as shown [here](https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/)

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -13,9 +13,10 @@ jobs:
         with:
           persist-credentials: false
       - name: Set up Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
         with:
           python-version: 3.9
+          cache: 'pip'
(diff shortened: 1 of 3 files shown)


isaacna:
I think I may have misread the purpose of the `cache` option but it seems that the cache is to be shared across steps, not job runs

isaacna:
I guess I could just go ahead and upgrade to `v2` but remove the `cache` option? @JacksonMaxfield thoughts on this?

JacksonMaxfield (approved):
Huh. Interesting. I would have hoped it was cached across runs too. Maybe this will still make an effect though because we install basic deps and then tox makes an entirely new env to run tests in so who knows. This may still be useful :shrug: :shrug: 

Thanks for upgrading!

PR #138, opened 2021-11-29
bugfix/host-videos-under-more-scenarios

JacksonMaxfield:
### Description of Changes

Major goal of this work is to host the video file provided to us from scraper under more scenarios.

However in addition to that, I added an option to the clean database function that allows the skipping of cleaning the index collections because those take a long time to fully delete. Additionally, I fix a bug I found while running the pipeline for king county which was that king county has some files that have names longer than 255 characters which Ubuntu doesn't like.

I can't seem to figure out why these changes will work in synchronous mode but not in parallel...

### DIFF
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -365,9 +365,6 @@ def convert_video_and_handle_host(
             save_name=f"{session_content_hash}-video.mp4",
         )
 
-        # Set session video_uri to uri in remote file store
-        session.video_uri = hosted_video_uri
-
         # Create fs to generate hosted media URL
         hosted_video_media_url = fs_functions.get_open_url_for_gcs_file(
             credentials_file=credentials_file,
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -36,6 +36,7 @@ log = logging.getLogger(__name__)
 
 class SessionProcessingResult(NamedTuple):
     session: Session
+    session_video_hosted_url: str
     audio_uri: str
     transcript: Transcript
     transcript_uri: str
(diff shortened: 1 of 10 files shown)


cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -301,33 +312,69 @@ def convert_video_to_mp4_and_upload(
...
-        # Return converted local mp4 filepath
-        return mp4_filepath
+        # Create fs to generate hosted media URL
+        hosted_video_media_url = fs_functions.get_open_url_for_gcs_file(
+            credentials_file=credentials_file,
+            uri=hosted_video_uri,
+        )
 
-    return video_filepath
+    return video_filepath, hosted_video_media_url
```
dphoria, line 377: Just wondering/trying to think in my head, does every code path to this statement include definition of the `hosted_video_media_url` somewhere in its path? In particular, I'm wondering about the `# Convert to mp4 if necessary` `if...else` blob.

dphoria (approved):
That one question/thought I commented earlier is it for me.

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: Yea I had to check this a number of times but it does exist in all cases.

Also, this is exactly why we use codecov! https://app.codecov.io/gh/CouncilDataProject/cdp-backend/compare/138/diff

image

Our tests hit every line of the code.

dphoria: Whoa this is pretty awesome. Cool!

JacksonMaxfield: Correction, looks like I am missing a single line 348, but that is simply setting `cdp_will_host` which we know will enable the hosting so no worries there.

dphoria: OK got it I follow

PR #144, opened 2021-12-19
bugfix/use-consistent-datetime-in-hash-generation

### LINKED ISSUE #142: Bug in DB model id generator / model hasher
See https://github.com/CouncilDataProject/cdp-frontend/issues/143
and https://github.com/CouncilDataProject/cdp-scrapers/issues/50

The hasher is sometimes producing multiple sessions for a single event. I believe this happens when the pipeline is running against the same event that has previously been processed.

I.e.
1. scrapper runs once (and completes processing like normal)
2. scrapper runs again and produces a different hash from the first run because the `Event` model is a reference (`ReferenceDocLoader`) and not an in-memory model

I can show this behavior by running the following:
```python
from cdp_backend.database import models as db_models
import fireo
from google.auth.credentials import AnonymousCredentials
from google.cloud.firestore import Client

from cdp_backend.database.functions import generate_and_attach_doc_hash_as_id

# Connect to the database
fireo.connection(client=Client(
    project="cdp-seattle-staging-dbengvtn",
    credentials=AnonymousCredentials()
))

def print_session_hashes(event_id):
    event = db_models.Event.collection.get(f"{db_models.Event.collection_name}/{event_id}")
    sessions = list(db_models.Session.collection.filter("event_ref", "==", event.key).fetch())
    for session in sessions:
        print(session.id)
        updated = generate_and_attach_doc_hash_as_id(session)
        print(updated.id)

print_session_hashes("c847010ce7a4")
# prints
# 71d1777f22c7
# 5626e734b33d
# e8b6fd84193f
# 5626e734b33d

print_session_hashes("9409f4d1ea09")
# prints
# 25d3a51447d0
# 9679fad34929
# 30701b4a0fbf
# 9679fad34929
```

Notice that in the above case, the second and fourth hashes match for each event.
This is because I am forcing it to rehash itself -- **but**, those second and fourth hashes match _neither_ of the original two hashes, which is bad. But since I am using the Python API to pull this data, it would seem to me that this may be happening because they are `ReferenceDocLoaders`... so something to explore next.

If I change this function to explicitely fetch the event object as an entire model (`db_models.Event`) in-memory rather than a `ReferenceDocLoader` object, it produces:

```python
def print_session_hashes(event_id):
    event = db_models.Event.collection.get(f"{db_models.Event.collection_name}/{event_id}")
    sessions = list(db_models.Session.collection.filter("event_ref", "==", event.key).fetch())
    for session in sessions:
        session.event_ref = session.event_ref.get()
        print(session.id)
        updated = generate_and_attach_doc_hash_as_id(session)
        print(updated.id)

print_session_hashes("c847010ce7a4")
# prints
# 71d1777f22c7
# e8b6fd84193f
# e8b6fd84193f
# e8b6fd84193f

print_session_hashes("9409f4d1ea09")
# prints
# 25d3a51447d0
# 30701b4a0fbf
# 30701b4a0fbf
# 30701b4a0fbf
```

Notice under this case that everything matches after the first hash. To my understanding this is because this is how the pipeline ran (simply due to insert order, the first time the pipeline ran is the third hash in the list, the second time the pipeline ran is the first hash in the list -- i.e. the order of the output hashes printed is most recent pipeline run produced hash, rehash after explicit model pull, original pipeline run produced hash, rehash after explicit model pull). The first time this event was processed, it had the model in memory, then the second time this event was processed (6 or so hours later), it had the Event as a reference and not in memory. If we force the model into memory, then it fixes itself.

I have looked at a ton of events and to be honest I am not sure how this doesn't effect every single event of ours. It seems to have only occurred for events in the Seattle Staging instance and for a couple of events.

I am still not sure if this is what caused the two variants of the issue as well. The scraper issue I originally labeled as "two of the same session for a single event". The frontend issue, @tohuynh found out that they were the same sessions but one was missing a transcript.

I truly do not know how deep this issue goes. I haven't found any King County event that fails.
I have found the following seattle staging events that fail:
* http://councildataproject.org/seattle-staging/#/events/c847010ce7a4 (duplicate session data)
* http://councildataproject.org/seattle-staging/#/events/9e03815fa280 (duplicate session data)
* http://councildataproject.org/seattle-staging/#/events/dbe0224f11a5 (missing transcript for second session)
* http://councildataproject.org/seattle-staging/#/events/da2b5f47e912 (missing transcript for second session)
* http://councildataproject.org/seattle-staging/#/events/c46c534afdce (missing transcript for second session)
* http://councildataproject.org/seattle-staging/#/events/9409f4d1ea09 (missing transcript for second session)
* http://councildataproject.org/seattle-staging/#/events/409e3e6f902f (missing transcript for second session)

cc @dphoria @tohuynh @isaacna

isaacna:
### Link to Relevant Issue

This pull request resolves #142 

### Description of Changes

When we hash with a `datetime` field, we keep timezone agnostic by converting the datetime to epoch millis. This keeps hashing consistent. I also refreshed the `field` variable after `setattr` just for sanity. 

Most of the background context can be found in the linked issue. 

### Testing:
You can see from the logs below that event hash is now consistent by searching for `FINAL HASH for <cdp_backend.database.models.Event object at`:
https://pastebin.com/dG8c3wRd 
- The print statements can be seen [here](https://github.com/CouncilDataProject/cdp-backend/commit/401c4da3204af840db511f7f1aa1a52cc2f6e3c6)

I also ran the following script to check that session is now consistent upon regeneration: 
```python
from cdp_backend.database import models as db_models
import fireo
from google.auth.credentials import AnonymousCredentials
from google.cloud.firestore import Client
from cdp_backend.database.functions import generate_and_attach_doc_hash_as_id

# Connect to the database
fireo.connection(client=Client(
    project="stg-cdp-seattle-isaac",
    credentials=AnonymousCredentials()
))

def print_session_hashes(event_id):
    event = db_models.Event.collection.get(f"{db_models.Event.collection_name}/{event_id}")
    print("EVENT ID: " + str(event.id))
    sessions = list(db_models.Session.collection.filter("event_ref", "==", event.key).fetch())
    for session in sessions:
        print("OLD: " + str(session.id))
        updated = generate_and_attach_doc_hash_as_id(session)
        print("UPDATED: " + str(updated.id))

print_session_hashes("6291946d4094")
"""
Prints:
EVENT ID: 6291946d4094
OLD: 7567518a5998
UPDATED: 7567518a5998

OLD: 97a4f821a3e7
UPDATED: 97a4f821a3e7

"""

```

### DIFF
--- a/cdp_backend/database/functions.py
+++ b/cdp_backend/database/functions.py
@@ -39,10 +39,7 @@ def generate_and_attach_doc_hash_as_id(db_model: Model) -> Model:
         The updated database model with the doc key set.
     """
     # Create hasher and hash primary values
-    print(db_model)
-    print("OG ID: " + str(db_model.id))
     hasher = sha256()
-    print(type(db_model))
     for pk in db_model._PRIMARY_KEYS:
         field = getattr(db_model, pk)
 
@@ -51,14 +48,10 @@ def generate_and_attach_doc_hash_as_id(db_model: Model) -> Model:
             # Ensure that the underlying model has an id
             # In place update to db_model for this field
 
-            print(type(field))
-            print("PRE GENERATE FIELD ID " + str(field.id))
             setattr(db_model, pk, generate_and_attach_doc_hash_as_id(field))
 
             # Update variable after setattr
             field = getattr(db_model, pk)
-            print(type(field))
-            print("GENERATED FIELD ID " + str(field.id))
 
             # Now attach the generated hash document path
             hasher.update(pickle.dumps(field.id, protocol=4))
@@ -67,22 +60,16 @@ def generate_and_attach_doc_hash_as_id(db_model: Model) -> Model:
         else:
             # Load the document for reference fields and add id to hasher
             if isinstance(field, ReferenceDocLoader):
-                print("REFERENCE DOC of " + str(type(field.get())))
-                print("HASHING REFERENCE WITH ID: " + str(field.get().id))
                 hasher.update(pickle.dumps(field.get().id, protocol=4))
             else:
+                # If datetime, hash with epoch millis to avoid timezone issues
                 if isinstance(field, datetime):
-                    print("A DATETIME")
                     field = field.timestamp()
-                    print("CONVERTING DATETIME TO " + str(field))
-                print("NORMAL CASE of " + str(type(field)))
-                print("HASHING " + pk + " "  + str(field))
                 hasher.update(pickle.dumps(field, protocol=4))
 
     # Set the id to the first twelve characters of hexdigest
     end_hash = hasher.hexdigest()[:12]
     db_model.id = end_hash
-    print("FINAL HASH for " + str(db_model) + str(end_hash))
 
     return db_model
 
--- a/cdp_backend/database/functions.py
+++ b/cdp_backend/database/functions.py
@@ -59,11 +59,13 @@ def generate_and_attach_doc_hash_as_id(db_model: Model) -> Model:
             # Now attach the generated hash document path
             hasher.update(pickle.dumps(field.id, protocol=4))
 
+        # If datetime, hash with epoch millis to avoid timezone issues
+        elif isinstance(field, datetime):
+            field = field.timestamp()
+            hasher.update(pickle.dumps(field, protocol=4))
+
         # Otherwise just simply add the primary key value
         else:
-            # If datetime, hash with epoch millis to avoid timezone issues
-            if isinstance(field, datetime):
-                field = field.timestamp()
             hasher.update(pickle.dumps(field, protocol=4))
 
     # Set the id to the first twelve characters of hexdigest
--- a/cdp_backend/database/functions.py
+++ b/cdp_backend/database/functions.py
@@ -53,13 +53,15 @@ def generate_and_attach_doc_hash_as_id(db_model: Model) -> Model:
 
             print(type(field))
             print("PRE GENERATE FIELD ID " + str(field.id))
-            new_field = generate_and_attach_doc_hash_as_id(field)
-            setattr(db_model, pk, new_field)
+            setattr(db_model, pk, generate_and_attach_doc_hash_as_id(field))
+
+            # Update variable after setattr
+            field = getattr(db_model, pk)
             print(type(field))
-            print("GENERATED FIELD ID " + str(new_field.id))
+            print("GENERATED FIELD ID " + str(field.id))
 
             # Now attach the generated hash document path
-            hasher.update(pickle.dumps(new_field.id, protocol=4))
+            hasher.update(pickle.dumps(field.id, protocol=4))
 
         # Otherwise just simply add the primary key value
         else:
--- a/cdp_backend/database/functions.py
+++ b/cdp_backend/database/functions.py
@@ -67,8 +67,7 @@ def generate_and_attach_doc_hash_as_id(db_model: Model) -> Model:
                 hasher.update(pickle.dumps(field, protocol=4))
 
     # Set the id to the first twelve characters of hexdigest
-    end_hash = hasher.hexdigest()[:12]
-    db_model.id = end_hash
+    db_model.id = hasher.hexdigest()[:12]
 
     return db_model
 
--- a/cdp_backend/database/functions.py
+++ b/cdp_backend/database/functions.py
@@ -39,7 +39,10 @@ def generate_and_attach_doc_hash_as_id(db_model: Model) -> Model:
         The updated database model with the doc key set.
     """
     # Create hasher and hash primary values
+    print(db_model)
+    print("OG ID: " + str(db_model.id))
     hasher = sha256()
+    print(type(db_model))
     for pk in db_model._PRIMARY_KEYS:
         field = getattr(db_model, pk)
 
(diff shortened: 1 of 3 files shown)


cdp_backend/database/functions.py
```diff
@@ -58,6 +61,9 @@ def generate_and_attach_doc_hash_as_id(db_model: Model) -> Model:
             if isinstance(field, ReferenceDocLoader):
                 hasher.update(pickle.dumps(field.get().id, protocol=4))
             else:
+                # If datetime, hash with epoch millis to avoid timezone issues
+                if isinstance(field, datetime):
```
JacksonMaxfield, line 65: Can you convert all of these to just top-level elifs like we had in the issue code pastes. This multi-level if -else if -else, etc is bad when they can all just be top level haha.

cdp_backend/database/functions.py
```diff
@@ -58,6 +61,9 @@ def generate_and_attach_doc_hash_as_id(db_model: Model) -> Model:
...
+                    field = field.timestamp()
```
JacksonMaxfield, line 66: Neat! Didn't know about this one.

JacksonMaxfield (commented):
I have a general comment. How should we handle the hashing changes generally. Like how do we ship this change to the existing instances. We don't have too much data yet and all the instances are controlled by us so we _could_ take the time to just reprocess all the data but that is rough. I feel like we should use this opportunity to figure out a decent method for shipping breaking changes like this, even if it's just a bugfix / patch.

A rough idea in my head is that we should have a bin script that is called like "process_cdp_upgrades" or something. This script does all the stuff we need it to from prior version to current version. In this specific case, it rehashes all of the database models that have datetimes in their primary keysets.

If all goes well and the script completes successfully, we store a key in the metadata collection in the database that is maybe: `"v3-0-0-patch": True` so that the next time the script runs it can check for this value and skip this process.

On the instance side we just have a github action that runs on push that runs this script?

isaacna:
> I have a general comment. How should we handle the hashing changes generally. Like how do we ship this change to the existing instances. We don't have too much data yet and all the instances are controlled by us so we _could_ take the time to just reprocess all the data but that is rough. I feel like we should use this opportunity to figure out a decent method for shipping breaking changes like this, even if it's just a bugfix / patch.
> 
> A rough idea in my head is that we should have a bin script that is called like "process_cdp_upgrades" or something. This script does all the stuff we need it to from prior version to current version. In this specific case, it rehashes all of the database models that have datetimes in their primary keysets.
> 

> 
> On the instance side we just have a github action that runs on push that runs this script?

I think this is a good idea for deploying breaking changes. 

> If all goes well and the script completes successfully, we store a key in the metadata collection in the database that is maybe: `"v3-0-0-patch": True` so that the next time the script runs it can check for this value and skip this process.

I like this method. I think we can also create a dictionary of patches to completion status in the metadata stack as well to consolidate all this info into one field.

For deploying this specific patch, I'm kinda hesitant to do try to tackle a partial reprocessing. Because it's a little trickier than just patching db models with datetime primary keys. It'd have to be anything that was recursively affected by this as well. This would affect multiple levels such as `Session` -> `Transcript` and other possible models?  

I'm personally more leaning towards a full reprocessing given our overall small data set, since a mistake with trying to do a partial db patch could leave it in a really odd/broken state

cdp_backend/database/functions.py
isaacna: Yep will do

cdp_backend/database/functions.py
isaacna: Yeah, I realized I supplied the datetime numerically as event data, so might as well just use that in the hashing

cdp_backend/database/functions.py
```diff
@@ -43,22 +43,28 @@ def generate_and_attach_doc_hash_as_id(db_model: Model) -> Model:
...
         else:
-            # Load the document for reference fields and add id to hasher
-            if isinstance(field, ReferenceDocLoader):
-                hasher.update(pickle.dumps(field.get().id, protocol=4))
-            else:
-                hasher.update(pickle.dumps(field, protocol=4))
+            # If datetime, hash with epoch millis to avoid timezone issues
+            if isinstance(field, datetime):
+                field = field.timestamp()
+            hasher.update(pickle.dumps(field, protocol=4))
```
JacksonMaxfield, line 67: Can I be nitpicky and say even this can be moved to top-level if, else?

JacksonMaxfield:
> I like this method. I think we can also create a dictionary of patches to completion status in the metadata stack as well to consolidate all this info into one field.
> 
> For deploying this specific patch, I'm kinda hesitant to do try to tackle a partial reprocessing. Because it's a little trickier than just patching db models with datetime primary keys. It'd have to be anything that was recursively affected by this as well. This would affect multiple levels such as `Session` -> `Transcript` and other possible models?
> 
> I'm personally more leaning towards a full reprocessing given our overall small data set, since a mistake with trying to do a partial db patch could leave it in a really odd/broken state

That's fair. I would say we should spend some time during a meeting talking about this problem. A dream would be able to run the script with params that say: "from v3.0.0 to v3.0.3" and it run all of the patches needed in order.

cdp_backend/database/functions.py
isaacna: Yeah sure, I can do move that as well

JacksonMaxfield (approved):
Merge whenever!

PR #147, opened 2022-01-06
bugfix/dont-fail-pipeline-on-bad-picture-urls

JacksonMaxfield:
### Description of Changes

This PR fixes or adds a feature (depending on how you look at it). We are experiencing some failures on the King County reprocessing: https://github.com/CouncilDataProject/king-county/runs/4729015978?check_suite_focus=true

This is due to the seat images failing to download, not because they have moved or don't exist but because our request has been denied by the host. This change makes it so that the pipeline will simply log the error and continue the rest of the pipeline. This imo is completely fine as seat and person images are optional anyway.

cc @dphoria because we probably want to update the king county seat images :joy:


isaacna (commented):
My bad for missing this but looks good to me! I agree that the images aren't necessary and can be backfilled later

PR #153, opened 2022-01-24
feature/truncate-sentences-in-transcript-str

### LINKED ISSUE #149: Change `Transcript` object repr / to string
I am not sure how to do this for a dataclass but I assume we can just do it in the object definition.

When I print out a `Transcript` object in a Jupyter notebook or any Python interpreter really, the `repr` and the tostring include the whole transcript details. I would prefer the shortened details:

```python
Transcript(
    generator='CDP WebVTT Conversion -- CDP v3.0.2',
    confidence=0.9699999999999663,
    session_datetime='2021-11-18T09:30:00-08:00',
    created_datetime='2022-01-12T05:05:00.633612',
    sentences=[...],
    annotations=None
)
```

Basically hiding the sentences because it makes the repr massive.

isaacna:
### Link to Relevant Issue

This pull request resolves #149 

### Description of Changes

Changing `Transcript`'s `__str__` method to truncate sentences. I did this with regex which was the least hacky thing I could think of. Alternatively, I was thinking about doing something like creating a deepcopy of `self` in `__str__` and replacing `sentences` with `[...]`, but that seems more hacky. There's not a standard way to hide/truncate a single property of a class afaik.

`repr()` will still show the full sentences.

### Testing
Running `print(EXAMPLE_TRANSCRIPT)`
Before:
```
Transcript(generator='JacksonGen -- Lib Version: 0.0.0', confidence=0.93325, session_datetime='2021-01-10T15:00:00', created_datetime='2022-01-24T02:05:41.708769', sentences=[Sentence(index=0, confidence=0.9, start_time=0.0, end_time=1.0, words=[Word(index=0, start_time=0.0, end_time=0.5, text='hello', annotations=None), Word(index=1, start_time=0.5, end_time=1.0, text='everyone', annotations=None)], text='Hello everyone.', speaker_index=0, speaker_name='Jackson Maxfield Brown', annotations=None), Sentence(index=1, confidence=0.95, start_time=1.0, end_time=2.0, words=[Word(index=0, start_time=1.0, end_time=1.5, text='hi', annotations=None), Word(index=1, start_time=1.5, end_time=2.0, text='all', annotations=None)], text='Hi all.', speaker_index=1, speaker_name='Isaac Na', annotations=None)], annotations=TranscriptAnnotations(sections=[SectionAnnotation(name='Call to Order', start_sentence_index=0, stop_sentence_index=2, generator='Jackson Maxfield Brown', description=None)]))
```

After
```
Transcript(generator='JacksonGen -- Lib Version: 0.0.0', confidence=0.93325, session_datetime='2021-01-10T15:00:00', created_datetime='2022-01-24T02:13:21.769205', sentences=[...], annotations=TranscriptAnnotations(sections=[SectionAnnotation(name='Call to Order', start_sentence_index=0, stop_sentence_index=2, generator='Jackson Maxfield Brown', description=None)]))
```

### DIFF
--- a/cdp_backend/pipeline/transcript_model.py
+++ b/cdp_backend/pipeline/transcript_model.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+import re
 from dataclasses import dataclass
 from datetime import datetime
 from typing import List, Optional
@@ -216,6 +217,13 @@ class Transcript:
     created_datetime: str
     sentences: List[Sentence]
     annotations: Optional[TranscriptAnnotations] = None
+
+    def __str__(self):
+        string = repr(self)
+        cleaned = re.sub(
+            "sentences=\[.*\], annotations", "sentences=[...], annotations", string
+        )
+        return cleaned
 
 
 ###############################################################################
--- a/cdp_backend/pipeline/transcript_model.py
+++ b/cdp_backend/pipeline/transcript_model.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-import re
 from dataclasses import dataclass
 from datetime import datetime
 from typing import List, Optional
@@ -218,12 +217,24 @@ class Transcript:
     sentences: List[Sentence]
     annotations: Optional[TranscriptAnnotations] = None
 
-    def __str__(self) -> str:
-        string = repr(self)
-        cleaned = re.sub(
-            r"sentences=\[.*\], annotations", "sentences=[...], annotations", string
-        )
-        return cleaned
+    def __repr__(self) -> str:
+        output = "Transcript("
+
+        # Use vars to maintain subclassing
+        for k, v in vars(self).items():
+            # Truncate sentences
+            if k == "sentences":
+                output += f"{k}=[...] (n={len(v)}), "
+
+            # Add quotes for strings
+            elif type(v) == str:
+                output += f"{k}='{v}', "
+
+            else:
+                output += f"{k}={v}, "
+
+        # Remove last comma and space and close parentheses
+        return output[:-2] + ")"
 
 
 ###############################################################################
--- a/cdp_backend/pipeline/transcript_model.py
+++ b/cdp_backend/pipeline/transcript_model.py
@@ -221,7 +221,7 @@ class Transcript:
     def __str__(self):
         string = repr(self)
         cleaned = re.sub(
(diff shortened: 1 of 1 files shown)


JacksonMaxfield:
I didn't even think of using a regex to strip. Cool!

Could you show what happens when you simply call the object? In ipython instead of using print just call it:
```python
t = read_some_transcript()
t
```

calling `t` like this calls the `__repr__` function. Idk if object standard is to use the `__str__` for `__repr__` if `__repr__` is undefined

isaacna:
This is a little tricky. In ipython, we get the non-truncated version so it must use `repr` if you just do `t`. 

Looks like `str` uses `repr` if [there's no implementation provided for the class](https://docs.python.org/3/library/stdtypes.html#str). So to change `repr` to truncate the string, that would mean we'd have to rewrite the logic to convert to a string itself, since that's what `repr` does in the first place.

I'm not familiar with Jupyter, but how are you using the transcript objects? According to [this post](https://stackoverflow.com/questions/18393701/the-difference-between-str-and-repr), `__str__` is supposed to be user friendly and `__repr__` is supposed to represent the object in more detail.

Based on this I think we can/should only use `str` when we want to truncate the objects

isaacna:
>  I think this has become the norm in the jupyter world because if Python is running in an ipython client (jupyter) it will actually call to a __repr_html__ function first if there is one. I.e. pandas dataframes, dask clusters, dask arrays, etc. all have HTML reprs that are quite useful.

> Anyway, its the norm in jupyter world to do that.

Gotcha, yeah I was mainly curious whether changing `repr` is something we _should_ do, but since that's the standard in Jupyter I think it's fine to do that.

> Technically all that is being printed is just dataclass_json.to_dict() so I think we could do something like this

Good idea! I did something similar except I used `vars` so we can preserve subclassing since `to_dict` was a pure json representation without class names. I also added `(n=...)` to the string generation

isaacna:
It will now output the following when just calling the object (same content as before):
```
Transcript(generator='JacksonGen -- Lib Version: 0.0.0', confidence=0.93325, session_datetime='2021-01-10T15:00:00', created_datetime='2022-01-24T21:30:40.583302', sentences=[...] (n=2), annotations=TranscriptAnnotations(sections=[SectionAnnotation(name='Call to Order', start_sentence_index=0, stop_sentence_index=2, generator='Jackson Maxfield Brown', description=None)]))
```

JacksonMaxfield (approved):
Wow. I have never even heard of the `vars` method. Very neat! Thanks!!

PR #155, opened 2022-01-26
feature/role-title-constants

dphoria:
### Link to Relevant Issue

This pull request is a prerequisite for resolving https://github.com/CouncilDataProject/cdp-scrapers/issues/66

### Description of Changes

Added `RoleTitle` string constants to `cdp_backend/database/constants.py`. This will be useful for `cdp_scrapers` when applying logic to ensure that a `Person` has only one `Role` with `title = RoleTitle.COUNCILMEMBER` per term.

I took `RoleTitle` members verbatim from Jackson's suggestions in this [comment](https://github.com/CouncilDataProject/cdp-scrapers/issues/66#issuecomment-1012578579). Certainly open to additions, modifications, etc.

For reference, below is the entire unique list of `OfficeRecordTitle` from all Legistar API `OfficeRecords` in Seattle.

```
Vice Chair
Member
Chair
Councilmember
City Clerk
None
Alternate
Mayor
Council President
```
(You gotta love `None` :rofl: )

For `OfficeRecordMemberType`,

```
Chair
Vice Chair
Member
```

### DIFF
--- a/cdp_backend/database/constants.py
+++ b/cdp_backend/database/constants.py
@@ -38,3 +38,12 @@ class MatterStatusDecision:
     ADOPTED = "Adopted"
     REJECTED = "Rejected"
     IN_PROGRESS = "In Progress"
+
+
+class RoleTitle:
+    councilmember = "Councilmember"
+    president = "President"
+    chair = "Chair"
+    vice_chair = "Vice Chair"
+    alternate = "Alternate"
+    member = "Member"
--- a/cdp_backend/database/constants.py
+++ b/cdp_backend/database/constants.py
@@ -42,7 +42,7 @@ class MatterStatusDecision:
 
 class RoleTitle:
     COUNCILMEMBER = "Councilmember"
-    PRESIDENT = "President"
+    COUNCILPRESIDENT = "Council President"
     CHAIR = "Chair"
     VICE_CHAIR = "Vice Chair"
     ALTERNATE = "Alternate"
--- a/cdp_backend/database/constants.py
+++ b/cdp_backend/database/constants.py
@@ -41,9 +41,9 @@ class MatterStatusDecision:
 
 
 class RoleTitle:
-    councilmember = "Councilmember"
-    president = "President"
-    chair = "Chair"
-    vice_chair = "Vice Chair"
-    alternate = "Alternate"
-    member = "Member"
+    COUNCILMEMBER = "Councilmember"
+    PRESIDENT = "President"
+    CHAIR = "Chair"
+    VICE_CHAIR = "Vice Chair"
+    ALTERNATE = "Alternate"
+    MEMBER = "Member"
--- a/cdp_backend/database/models.py
+++ b/cdp_backend/database/models.py
@@ -189,9 +189,7 @@ class Role(Model):
 
     id = fields.IDField()
     title = fields.TextField(
-        validator=validators.create_constant_value_validator(
-            RoleTitle, True
-        )
+        validator=validators.create_constant_value_validator(RoleTitle, True)
     )
     person_ref = fields.ReferenceField(Person, required=True, auto_load=False)
     body_ref = fields.ReferenceField(Body, auto_load=False)
--- a/cdp_backend/database/models.py
+++ b/cdp_backend/database/models.py
@@ -204,7 +204,7 @@ class Role(Model):
     @classmethod
     def Example(cls) -> Model:
         role = cls()
-        role.title = RoleTitle.PRESIDENT
+        role.title = RoleTitle.COUNCILPRESIDENT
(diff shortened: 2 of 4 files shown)


cdp_backend/database/constants.py
```diff
@@ -38,3 +38,12 @@ class MatterStatusDecision:
     ADOPTED = "Adopted"
     REJECTED = "Rejected"
     IN_PROGRESS = "In Progress"
+
+
+class RoleTitle:
+    COUNCILMEMBER = "Councilmember"
+    PRESIDENT = "President"
```
tohuynh, line 45: Change to "Council President"?

tohuynh, line 45: Change to "Council President"?

dphoria: I think that would be good, as long as we know that the concept/title of president is used only for city council. i.e. Municipalities do not use "president" for something like the "Public Transportation Committee", etc. I guess not???

tohuynh:
Also, don't know if you want to add it to this PR, but a validator needs to be added to https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/models.py#L190.

See an example of a constant validator: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/models.py#L612

dphoria:
> Also, don't know if you want to add it to this PR, but a validator needs to be added to https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/models.py#L190.
> 
> See an example of a constant validator: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/models.py#L612

Ohhh thank you so much. I'm switching this PR to draft for now then. :sweat_smile:

cdp_backend/database/constants.py
tohuynh: Not sure too. I think they use "Chair" for that?

dphoria: Yeah. At least for Seattle, the only title (in Legistar) with the word "president" in it is "council president." So perhaps it is safe to define/use "Council President", only.

JacksonMaxfield (approved):
Looks good to me! Thanks! Will let one of the front end folks approve too before merge

isaacna (approved):
Looks good to me!

cdp_backend/database/models.py
```diff
@@ -187,7 +188,9 @@ class Role(Model):
     """
 
     id = fields.IDField()
-    title = fields.TextField(required=True)
+    title = fields.TextField(
+        validator=validators.create_constant_value_validator(RoleTitle, True)
+    )
```
tohuynh, line 193: I notice that the other required constant field like [`matterstatus.status`](https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/models.py#L616) have a `required=True`, while this one doesn't. It doesn't matter that it is missing right, as long the second arg to `create_constant_value_validator` is `True`?

cc @isaacna @JacksonMaxfield

dphoria: Oops that's my mistake. Thank you for catching that!

JacksonMaxfield: Yea this is just a styling thing. The second arg of `True` will be used for the `required` kwarg. Can include or not.

The only reason why I commonly include the full kwarg name is because if we ever change the API, and we shift the order of the arguments, the kwargs still get filled properly regardless of the order. But I don't think we plan on changing the arg order haha. Either way, I'm not concerned.

dphoria: I should still probably changed it so set `required` for `fields.TextField()`, i.e.

```Python
title = fields.TextField(
    required=True, # this one
    validator=validators.create_constant_value_validator(RoleTitle, True),
)
```

isaacna: I had the same thought earlier but also agree that keeping the arg name is not a big deal. I think we can add them for Jackson's reasons above and also readability. Earlier I was gonna comment why we changed `title` to not be required, but then I saw that `True` was for `is_required`. If we add the arg name it becomes more clear

isaacna: I think if we set the FireO arg `required`, than we can just get rid of the `is_required` field in `create_constant_value_validator`. Same goes for everywhere else since I don't think it really adds any functionality

tohuynh: Isaac's suggestion makes sense, we only need one way to check if something is required.

dphoria: I'll change as above, but I guess this is food for thought for the future then.

JacksonMaxfield: Agree. Change to match others for now. Can you open an issue to track the above thoughts as well?

PR #157, opened 2022-01-27
bug/catch-timeout-in-event-gather

### LINKED ISSUE #154: Catch timeout / request rejections during event archiving
### Describe the Bug

When we are storing links to the matter attachments during the event ingestion pipeline `store_event_processing_result` task, we run existance checks on every supporting file. Because there can be a lot of these supporting files, we occasionally encounter request rejections for hitting the host too often / too quickly.

### Expected Behavior

_What did you expect to happen instead?_

The file should be skipped. Simply catching the exception and moving on instead of failing the pipeline.

### Reproduction

_Steps to reproduce the behavior and/or a minimal example that exhibits the behavior._

https://github.com/CouncilDataProject/seattle-staging/runs/4930903577?check_suite_focus=true

### Environment

_Any additional information about your environment._

Ubuntu 20.04 and cdp-backend v3.0.3 -- see GitHub Actions details.

isaacna:
### Link to Relevant Issue

This pull request resolves #154 

### Description of Changes

The [error here](https://github.com/CouncilDataProject/seattle-staging/runs/4930903577?check_suite_focus=true) was caused by a `requests.ConnectionError`. 

I added a catch for `ConnectionError` for every `try` statement that uploads a db model and has `resource_exists` validation. 

I'm assuming that wherever we don't have a `try` statement, we wouldn't want to proceed in the pipeline if something errors out during that `upload_db_model` step. But let me know if this assumption isn't correct

### DIFF
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -14,6 +14,7 @@ from fsspec.core import url_to_fs
 from gcsfs import GCSFileSystem
 from prefect import Flow, task
 from prefect.tasks.control_flow import case, merge
+from requests import ConnectionError
 
 from ..database import constants as db_constants
 from ..database import functions as db_functions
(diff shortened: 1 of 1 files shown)


JacksonMaxfield (approved):
Looks good! Thanks for making these changes so quickly!

Issue #156, opened 2022-01-26
Store content hash string on the session db model


JacksonMaxfield:
### Feature Description

_A clear and concise description of the feature you're requesting._

Currently if we want to look up the content hash for a session we need to go to through any transcript for the session, then the transcripts file ref, then do a split on "-" and take the first part of the split.

### Use Case

_Please provide a use case to help us understand your request in context._

Would make it much faster to get the audio file for a session if we stored the content hash on the session model itself.

### Solution

_Please describe your ideal solution._

Update the db model and the pipeline to store the content hash. Also write a script that updates old sessions with their session hashes using the transcript file ref to split process as described above.

### Alternatives

_Please describe any alternatives you've considered, even if you've dismissed them._

isaacna:
Linking #159 which created a script to update an old db instance. Last remaining part is to create a github action to run the script

JacksonMaxfield:
Real quick. I thought about this recently. I think maybe we simply make a github action that allows maintainers the ability to run any arbitrary bin script?

In this case, the PR you just linked added a bin script to the package called: `add_content_hash_to_sessions`

If we had a github action that simply installed cdp-backend, loaded the secrets / env, and was parametrized with a bin script, we could make the github action just accept the bin script name as a param.

JacksonMaxfield:
Basically, my thinking here is instead of making a complicated "checker" for version, we leave it to the maintainer to run the script?

isaacna:
> If we had a github action that simply installed cdp-backend, loaded the secrets / env, and was parametrized with a bin script, we could make the github action just accept the bin script name as a param.

I'm good with this idea, but is there an easy way to handle scripts with different arguments? I guess we could always use env variables, since I didn't see anything Github Action's documentation for passing in params without hardcoding them.

> Basically, my thinking here is instead of making a complicated "checker" for version, we leave it to the maintainer to run the script?

Yeah I don't think there's necessarily a need for a version checker, but maybe there's some other way we can handle a scenario where some script/fix _needs_ to be ran in order for updates to be pulled in. But ideally we shouldn't need to do this unless we have another data backfill situation like the hashing bug

JacksonMaxfield:
we could simply allow the full command to be passed through.

```bash
some_fake_cdp_bin_script --arg1=2 --arg2="hello"
```

JacksonMaxfield:
Going to move this issue over to cookiecutter for tracking there

PR #158, opened 2022-01-28
feature/store-session-content-hash-in-session-db-model

isaacna:
### Link to Relevant Issue

This pull request resolves first part of #156 

### Description of Changes

Added `session_content_hash` to the db model and all relevant places in the event gather pipeline that are necessary to support that.

Will write a script later that retroactively does this for existing session db rows


JacksonMaxfield (approved):
Looks great! Thanks!

PR #159, opened 2022-01-29
feature/script-for-retroactive-session-content-hashing

isaacna:
### Link to Relevant Issue

This pull request resolves #158 

### Description of Changes

Added a script for adding the content hash for every session in a firestore instance. 

Running this script would be:
```
add_content_hash_to_sessions --google_credentials_file {PATH_TO_CREDS_FILE}
```

### Testing

Ran this on my personal firestore instance and tested the following:
- All session content hashes are added
- We properly log which sessions still don't have the content hashes

### DIFF
--- a/cdp_backend/bin/add_content_hash_to_sessions.py
+++ b/cdp_backend/bin/add_content_hash_to_sessions.py
@@ -9,7 +9,6 @@ from pathlib import Path
 
 import fireo
 
-from cdp_backend.database import models as db_models
 from cdp_backend.database.models import Session, Transcript
 
 ###############################################################################
@@ -30,7 +29,8 @@ class Args(argparse.Namespace):
     def __parse(self) -> None:
         p = argparse.ArgumentParser(
             prog="add_content_hash_to_sessions",
-            description="Add content hash to all existing session rows in the database.",
+            description="Add content hash to all existing session rows in "
+            + "the database.",
         )
         p.add_argument(
             "--google_credentials_file",
@@ -96,10 +96,12 @@ def add_content_hash_to_sessions(google_creds_path: Path) -> None:
             unfixed_sessions.remove(session.id)
 
     # Log any sessions that still don't have a content hash
-    # Could happen if there are db inconsistencies (like a session is orphaned w/o a transcript)
+    # Could happen if there are db inconsistencies
+    # (like a session is orphaned w/o a transcript)
     if unfixed_sessions:
         log.error(
-            f"The following sessions were not fixed with a content hash: {unfixed_sessions}"
+            "The following sessions were not fixed with a "
+            + f"content hash: {unfixed_sessions}"
         )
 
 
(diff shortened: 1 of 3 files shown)


cdp_backend/bin/add_content_hash_to_sessions.py
```diff
@@ -0,0 +1,120 @@
...
+
+class Args(argparse.Namespace):
+    def __init__(self) -> None:
+        self.__parse()
+
+    def __parse(self) -> None:
+        p = argparse.ArgumentParser(
+            prog="add_content_hash_to_sessions",
+            description="Add content hash to all existing session rows in "
+            + "the database.",
```
JacksonMaxfield, line 33: Super nitpick, can you change this to:
```python
(
    "Add content hash to all existing session rows in "
    "the database."
)
```

Strings dont need addition if they simply follow each other.

cdp_backend/bin/add_content_hash_to_sessions.py
```diff
@@ -0,0 +1,120 @@
...
+
+        # If no session_content_hash found
+        if not session.session_content_hash:
+            # Extract hash  from file
+            session_content_hash = _file.name.split("-")[0]
+
+            # Need to set event reference since autoload is disabled,
+            # and FireO throws an error if the model has a ReferenceDocLoader
+            # on a property during any db write action
+            session.event_ref = session.event_ref.get()
```
JacksonMaxfield, line 79: this is wild

JacksonMaxfield (approved):
I think this makes sense for the most part! I will try to come up with a deployment strategy for implementing in cookiecutter. I kind of want to make a GitHub action that runs this and any future changes we need to make.

"Action: Upgrade 3.0.3 - 3.0.4 -- click to run" if that makes sense?

isaacna:
> I kind of want to make a GitHub action that runs this and any future changes we need to make.

> "Action: Upgrade 3.0.3 - 3.0.4 -- click to run" if that makes sense?

I'm good with this. Not sure actions supports this, but we could try to make this action appear conditionally on if the version is <= 3.0.3

cdp_backend/bin/add_content_hash_to_sessions.py
isaacna: Will do, but had to add a mypy ignore since it reads the type as a tuple

cdp_backend/bin/add_content_hash_to_sessions.py
isaacna: Yeah there were a few more bugs with FireO that made writing this a lot more difficult than it should have been

cdp_backend/bin/add_content_hash_to_sessions.py
JacksonMaxfield: uhhhhhhhhh dont add a comma. it shouldn't

cdp_backend/bin/add_content_hash_to_sessions.py
```diff
@@ -0,0 +1,122 @@
...
+class Args(argparse.Namespace):
+    def __init__(self) -> None:
+        self.__parse()
+
+    def __parse(self) -> None:
+        p = argparse.ArgumentParser(
+            prog="add_content_hash_to_sessions",
+            description=(
+                "Add content hash to all existing session rows "  # type: ignore
+                "in the database.",
```
JacksonMaxfield, line 34: if you remove this trailing comma, mypy shouldnt see this as a tuple

PR #35, opened 2020-12-23
feature/add-matter-status-constants

tohuynh:
**Pull request recommendations:**
- [x] Name your pull request _your-development-type/short-description_. Ex: _feature/read-tiff-files_
- [ ] Link to any relevant issue in the PR description. Ex: _Resolves [gh-12], adds tiff file format support_
- [x] Provide context of changes.
- Add useful matter status constants for `MatterStatus` database model. The constants are `Adopted`, `Rejected`, and `In Progress`.
- Provide validators for them in the database models. Instead of creating a validator for each constant class, I created a factory function that creates the validator to avoid duplication.
- Also added `status` to the `Matter` ingestion model. I decided to put it there to follow the `sponsors` example. Similar to how `sponsors` will be used to create/update `MatterSponsor`, `status` will be used to create/update `MatterStatus` in the pipeline.
- [x] Provide relevant tests for your feature or bug fix.
- [ ] Provide or update documentation for any feature added by your pull request.

Thanks for contributing!

PR #163, opened 2022-02-10
feature/remove-is-required-from-constant-validator

### LINKED ISSUE #160: Remove `is_required` from `create_constant_value_validator`
### Feature Description

As discussed in #155 

We don't need `is_required` in the function `create_constant_value_validator`. We can just use the `required` arg that is built in for all FireO fields. We should just remove the param in `create_constant_value_validator` and adjust the usages to keep things consistent

isaacna:
### Link to Relevant Issue

This pull request resolves #160 

### Description of Changes

Removing `is_required` and since we can use FireO's `Required=True` wherever applicable.

### DIFF
--- a/cdp_backend/tests/database/test_validators.py
+++ b/cdp_backend/tests/database/test_validators.py
@@ -136,13 +136,12 @@ def test_remote_resource_exists(
 @pytest.mark.parametrize(
     "decision, expected_result",
     [
-        (None, False),
         ("Approve", True),
         ("INVALID", False),
     ],
 )
 def test_vote_decision_is_valid(decision: str, expected_result: bool) -> None:
-    validator_func = validators.create_constant_value_validator(VoteDecision, True)
+    validator_func = validators.create_constant_value_validator(VoteDecision)
     actual_result = validator_func(decision)
     assert actual_result == expected_result
 
@@ -150,7 +149,6 @@ def test_vote_decision_is_valid(decision: str, expected_result: bool) -> None:
 @pytest.mark.parametrize(
     "decision, expected_result",
     [
-        (None, True),
         ("Passed", True),
         ("INVALID", False),
     ],
@@ -159,7 +157,7 @@ def test_event_minutes_item_decision_is_valid(
     decision: str, expected_result: bool
 ) -> None:
     validator_func = validators.create_constant_value_validator(
-        EventMinutesItemDecision, False
+        EventMinutesItemDecision
     )
     actual_result = validator_func(decision)
     assert actual_result == expected_result
@@ -168,15 +166,12 @@ def test_event_minutes_item_decision_is_valid(
 @pytest.mark.parametrize(
     "decision, expected_result",
     [
-        (None, False),
         ("Adopted", True),
         ("INVALID", False),
     ],
 )
 def test_matter_status_decision_is_valid(decision: str, expected_result: bool) -> None:
-    validator_func = validators.create_constant_value_validator(
-        MatterStatusDecision, True
-    )
+    validator_func = validators.create_constant_value_validator(MatterStatusDecision)
     actual_result = validator_func(decision)
     assert actual_result == expected_result
 
@@ -184,12 +179,11 @@ def test_matter_status_decision_is_valid(decision: str, expected_result: bool) -
 @pytest.mark.parametrize(
     "title, expected_result",
     [
-        (None, False),
         ("Councilmember", True),
         ("INVALID", False),
     ],
 )
 def test_role_title_is_valid(title: str, expected_result: bool) -> None:
-    validator_func = validators.create_constant_value_validator(RoleTitle, True)
+    validator_func = validators.create_constant_value_validator(RoleTitle)
     actual_result = validator_func(title)
     assert actual_result == expected_result
--- a/cdp_backend/database/models.py
+++ b/cdp_backend/database/models.py
@@ -190,7 +190,7 @@ class Role(Model):
     id = fields.IDField()
     title = fields.TextField(
         required=True,
-        validator=validators.create_constant_value_validator(RoleTitle, True),
+        validator=validators.create_constant_value_validator(RoleTitle),
     )
     person_ref = fields.ReferenceField(Person, required=True, auto_load=False)
     body_ref = fields.ReferenceField(Body, auto_load=False)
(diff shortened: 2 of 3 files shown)


cdp_backend/tests/database/test_validators.py
```diff
@@ -136,21 +136,19 @@ def test_remote_resource_exists(
 @pytest.mark.parametrize(
     "decision, expected_result",
     [
-        (None, False),
```
JacksonMaxfield, line 139: This is making me question what the pattern is. Before I approve can you show me code that has a random FireO model with a field thats required fail to push with a `None` value?

Like I want to make sure that regardless of _this_ change, that `None` values for required fields are still rejected from upload.

isaacna: Sure, here's an example code snippet I ran:
```python
matter_status = MatterStatus()
matter_status.matter_ref = Matter.Example()
matter_status.status = None
upload_db_model(matter_status, config.google_credentials_file)
```

It will raise a FireO `RequiredField` exception:
```
Traceback (most recent call last):
  File "/Users/Isaac/Desktop/CDP/cdp-backend/cdp_backend/database/functions.py", line 107, in upload_db_model
    db_model = db_model.upsert(transaction=transaction, batch=batch)
  File "/Users/Isaac/.local/share/virtualenvs/cdp-backend-9V6eR1KV/lib/python3.7/site-packages/fireo/models/model.py", line 352, in upsert
    return self.save(transaction=transaction, batch=batch, merge=True)
  File "/Users/Isaac/.local/share/virtualenvs/cdp-backend-9V6eR1KV/lib/python3.7/site-packages/fireo/models/model.py", line 346, in save
    return self.__class__.collection.create(self, transaction, batch, merge, no_return, **self._get_fields())
  File "/Users/Isaac/.local/share/virtualenvs/cdp-backend-9V6eR1KV/lib/python3.7/site-packages/fireo/managers/managers.py", line 230, in create
    return self.queryset.create(mutable_instance, transaction, batch, merge, no_return, **field_list)
  File "/Users/Isaac/.local/share/virtualenvs/cdp-backend-9V6eR1KV/lib/python3.7/site-packages/fireo/queries/query_set.py", line 56, in create
    return CreateQuery(self.model_cls, mutable_instance, no_return, **kwargs).exec(transaction_or_batch, merge)
  File "/Users/Isaac/.local/share/virtualenvs/cdp-backend-9V6eR1KV/lib/python3.7/site-packages/fireo/queries/create_query.py", line 158, in exec
    return query_wrapper.ModelWrapper.from_query_result(self.model, self._raw_exec(merge=merge))
  File "/Users/Isaac/.local/share/virtualenvs/cdp-backend-9V6eR1KV/lib/python3.7/site-packages/fireo/queries/create_query.py", line 143, in _raw_exec
    ref.set(self._parse_field(), merge=merge)
  File "/Users/Isaac/.local/share/virtualenvs/cdp-backend-9V6eR1KV/lib/python3.7/site-packages/fireo/queries/create_query.py", line 97, in _parse_field
    fv = f.get_value(self.query.get(f.name))
  File "/Users/Isaac/.local/share/virtualenvs/cdp-backend-9V6eR1KV/lib/python3.7/site-packages/fireo/fields/base_field.py", line 122, in get_value
    val = self.field_attribute.parse(val, ignore_required, ignore_default)
  File "/Users/Isaac/.local/share/virtualenvs/cdp-backend-9V6eR1KV/lib/python3.7/site-packages/fireo/fields/field_attribute.py", line 86, in parse
    raise RequiredField(f'"{self.field.__class__.__name__}" is required for model {self.field.model_cls} '
fireo.fields.errors.RequiredField: "TextField" is required for model <class 'cdp_backend.database.models.MatterStatus'> but received no default and no value.
```

Same thing happens when we don't set `status` at all

isaacna: If upload fails due to a required field not being present, do we want to fail on the pipeline or keep proceeding? If it's the latter, then I can update the PR to catch `RequiredField` for uploads

JacksonMaxfield: Ahhh I think it depends on the case. I wonder if this was why we split them up initially? I will go through this PR and see which ones are _needed_ and which one are truly optional

isaacna:
I don't think To had a specific reason, but may have just added it if he didn't know about the `required` FireO field: 
https://github.com/CouncilDataProject/cdp-backend/pull/35 

> Looking at the pipeline code: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/event_gather_pipeline.py#L1133
>
> If we were to merge this PR, that line wouldn't break it would just raise a different error correct?

Yeah this wouldn't break, it'd just throw `RequiredField` instead of `FieldValidationFailed`.

 > BUT, looking at this: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/event_gather_pipeline.py#L1413
>
> I think this would break some stuff because passing None would now be allowed through the validator I think but would raise a requiredField error that isn't current caught?

So I tested this out, and technically there would be no change to the existing functionality. If you look at [this PR](https://github.com/CouncilDataProject/cdp-backend/commit/8743498b4a7c2814a0fe9febe0bec92977a4c921). We used the FireO `required` field before, but even with Sung's changes, the FireO `required` check happens before the validation check for a given field. 

So we're already (and always have been) throwing `RequiredField` if it's missing or `None`. If we want to pass instead of erroring out we'd have to catch `RequiredField` as well if `None` is passed. But I wasn't sure if we'd necessarily want to because if the error is `RequiredField` then that means the ingestion data is wrong 

Btw, the `FieldValidationFailed` and `RequiredField` errors on FireO models aren't triggered until we attempt to perform a write operation with the db model

isaacna:
So TLDR: the validator param `is_required` isn't even doing anything, we should be able to just remove it

JacksonMaxfield:
> So I tested this out, and technically there would be no change to the existing functionality. If you look at [this PR](https://github.com/CouncilDataProject/cdp-backend/commit/8743498b4a7c2814a0fe9febe0bec92977a4c921). We used the FireO `required` field before, but even with Sung's changes, the FireO `required` check executes before the validation check for a given field.

Oh. Well that alone has convinced me. Sounds good. Thanks for the additional comments and research.

JacksonMaxfield (approved):

PR #164, opened 2022-02-16
bugfix/revert-validator-fixes

JacksonMaxfield:
Reverts CouncilDataProject/cdp-backend#163

This PR created pipeline problems: https://github.com/CouncilDataProject/seattle-staging/runs/5219968602?check_suite_focus=true

Relevant part of the pipeline: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/event_gather_pipeline.py#L1373

Logged warning:
```
[WARNING: event_gather_pipeline:1389 2022-02-16 17:04:24,396] Provided 'decision' is not an approved constant. Provided: 'None' Should be one of: ['Failed', 'Passed'] See: cdp_backend.database.constants.EventMinutesItemDecision. Creating EventMinutesItem without decision value.
```

DB model: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/models.py#L556

The only field on that db model with a validator is `decision` but it's not a required field, so why didn't this go through?

Because the validator _function_ doesn't allow `None` as an option. This is why I was so confused.

I am going to revert this change _and_ change the parameter name to be more explicit about what that parameter is doing. It isn't making it `required` or not, it is allowing `None` as a valid option.

I will also update tests and add more docs to make sure we don't forget this again :joy:

### DIFF
--- a/cdp_backend/database/validators.py
+++ b/cdp_backend/database/validators.py
@@ -157,7 +157,7 @@ def resource_exists(uri: Optional[str], **kwargs: str) -> bool:
 
 
 def create_constant_value_validator(
-    constant_cls: Type, is_required: bool
+    constant_cls: Type, allow_none: bool
 ) -> Callable[[str], bool]:
     """
     Create a validator func that validates a value is one of the valid values.
@@ -166,13 +166,18 @@ def create_constant_value_validator(
     ----------
     constant_cls: Type
         The constant class that contains the valid values.
-    is_required: bool
-        Whether the value is required.
+    allow_none: bool
+        Whether to allow `None` as a valid value for this field.
 
     Returns
     -------
     validator_func: Callable[[str], bool]
         The validator func.
+
+    Notes
+    -----
+    If the field the created validation function is attached to is optional,
+    (i.e. required=True, is not set) `allow_none` should be set to True.
     """
 
     def is_valid(value: str) -> bool:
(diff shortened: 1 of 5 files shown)


cdp_backend/database/validators.py
```diff
@@ -156,19 +156,28 @@ def resource_exists(uri: Optional[str], **kwargs: str) -> bool:
...
+) -> Callable[[str], bool]:
     """
     Create a validator func that validates a value is one of the valid values.
 
     Parameters
     ----------
     constant_cls: Type
         The constant class that contains the valid values.
+    allow_none: bool
+        Whether to allow `None` as a valid value for this field.
```
isaacna, line 170: I don't think we actually need this field at all, since this will basically always be the inverse of `required` right? 

If FireO `required` is True, then that should catch the `None` case. If we don't set FireO `required`, then `allow_none` should be true.

Can we just do the following, and let `required` handle when `None` is not allowed?

```python
return value is None or value in allowed_attrs
```

isaacna (changes requested):
This makes sense, thanks for discovering that lost piece of context as to why the function was originally built that way. 

My one comment is I think we could actually remove `allow_none` completely

cdp_backend/database/validators.py
JacksonMaxfield: I see your point! That makes a lot of sense! Will give it a go

tohuynh:
I didn't think my validator function would cause such a headache!

Given Isaac's comment, I think you only need a `None` test case for when the data model field is optional, not for when the data model field is required since that test case is covered (hopefully) by FireO.

JacksonMaxfield:
Should be good to go now!

isaacna (approved):
Looks good to me, thanks for making the changes!

tohuynh (approved):

PR #165, opened 2022-02-17
bugfix/clean-pictograms-from-transcripts-before-indexing

### LINKED ISSUE #151: Context span selection fails during index creation
### Describe the Bug

https://github.com/CouncilDataProject/seattle-staging/runs/4862736150?check_suite_focus=true

Looks like the `closest_term` was `""` and an empty string wasn't found in the list. Regardless of how valuable that is. This was from seattle-staging and we can run and store the index locally so hopefully this will allow us to check out both what term is being processed and what is going on in general.

JacksonMaxfield:
### Link to Relevant Issue

This pull request resolves #151 

### Description of Changes

_Finally_ found time to fix this one. Can't believe the bug either.

Looks like a transcript from Seattle has a `♪` pictogram / emoticon in it...
[link -- start at 30:09 -- or search for `♪`](http://councildataproject.org/seattle-staging/#/events/c5d6d360b88d?s=0&t=1809)

This fixes the pipeline by adding a function to clean all common pictograms / emojis from the sentence before stemming and fuzzy matching for context spans.

Tested by running the pipeline and storing the index locally:
`run_cdp_event_index -n 1 --store_local --parallel ../configs-and-special-events/seattle.json`

### DIFF
--- a/cdp_backend/utils/string_utils.py
+++ b/cdp_backend/utils/string_utils.py
@@ -107,10 +107,7 @@ def clean_text(
     # Remove gaps in string
     try:
         cleaned_doc = re.sub(r" {2,}", " ", cleaned_text)
-        if cleaned_doc[0] == " ":
-            cleaned_doc = cleaned_doc[1:]
-        if cleaned_doc[-1] == " ":
-            cleaned_doc = cleaned_doc[:-1]
+        cleaned_doc = cleaned_doc.strip()
 
     # IndexError occurs when the string was cleaned and it contained entirely stop
     # words or punctuation for some reason
--- a/cdp_backend/utils/string_utils.py
+++ b/cdp_backend/utils/string_utils.py
@@ -12,7 +12,38 @@ log = logging.getLogger(__name__)
 ###############################################################################
 
 
-def clean_text(text: str, clean_stop_words: bool = False) -> str:
+def remove_emojis(text: str) -> str:
+    """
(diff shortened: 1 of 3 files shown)


JacksonMaxfield:
> Looks good to me! Honestly pretty suprised that Google speech-to-text generated an emoji haha

I think this transcript is from a converted closed caption. Which makes a bit more sense 😂

cdp_backend/utils/string_utils.py
```diff
@@ -57,18 +91,22 @@ def clean_text(text: str, clean_stop_words: bool = False) -> str:
...
+    if clean_emojis:
+        cleaned_text = remove_emojis(cleaned_text)
 
     # Remove gaps in string
     try:
-        cleaned_doc = re.sub(r" {2,}", " ", cleaned_stopwords)
+        cleaned_doc = re.sub(r" {2,}", " ", cleaned_text)
         if cleaned_doc[0] == " ":
             cleaned_doc = cleaned_doc[1:]
         if cleaned_doc[-1] == " ":
```
tohuynh, line 112: Isn't this `.strip()`. That's more readable to me.

tohuynh (approved):
Looks good

cdp_backend/utils/string_utils.py
JacksonMaxfield: I think you're correct. :facepalm:

PR #169, opened 2022-02-21
feature/downloading-youtube-videos

### LINKED ISSUE #161: Add a function for downloading youtube videos
### Feature Description

[Context](https://github.com/CouncilDataProject/cdp-scrapers/pull/72)

Add a function for downloading youtube videos

### Use Case

Some deployments (such as portland) may upload their videos via youtube. We need to be able to download the video for both audio stripping for session hash generation, trasncription, and speaker classification

### Solution

We can use [youtube-dl](https://github.com/ytdl-org/youtube-dl/blob/master/README.md#embedding-youtube-dl)

Shak2000:
### Link to Relevant Issue

This pull request resolves #161 

### Description of Changes

Make it possible for the program to download a YouTube video embedded onto a website. This is needed for the Portland scraper.

### DIFF
--- a/cdp_backend/tests/utils/test_file_utils.py
+++ b/cdp_backend/tests/utils/test_file_utils.py
@@ -254,5 +254,7 @@ def test_youtube_downloader(
     actual_uri = file_utils.resource_copy(youtube_uri, resources_dir, True)
     expected_uri = str(resources_dir / expected)
     assert actual_uri == expected_uri
+    assert Path(actual_uri).exists()
+    assert Path(actual_uri).is_file()
 
     os.remove(actual_uri)
--- a/cdp_backend/tests/utils/test_file_utils.py
+++ b/cdp_backend/tests/utils/test_file_utils.py
@@ -19,11 +19,14 @@ from cdp_backend.utils.file_utils import (
     resource_copy,
 )
 
+from .. import test_utils
 from ..conftest import (
     EXAMPLE_MKV_VIDEO_FILENAME,
     EXAMPLE_VIDEO_FILENAME,
     EXAMPLE_VIDEO_HD_FILENAME,
-    EXAMPLE_YOUTUBE_VIDEO,
+    EXAMPLE_YOUTUBE_VIDEO_EMBEDDED,
+    EXAMPLE_YOUTUBE_VIDEO_PARAMETER,
+    EXAMPLE_YOUTUBE_VIDEO_SHORT,
 )
 
 #############################################################################
@@ -212,6 +215,10 @@ def test_hover_thumbnail_generator(
     os.remove(result)
 
 
+@pytest.mark.skipif(
+    not test_utils.internet_is_available(),
+    reason="No internet connection",
+)
 @pytest.mark.parametrize(
     "video_uri, expected",
     [
@@ -227,10 +234,16 @@ def test_convert_video_to_mp4(
     assert file_utils.convert_video_to_mp4(filepath) == str(resources_dir / expected)
 
 
+@pytest.mark.skipif(
+    not test_utils.internet_is_available(),
+    reason="No internet connection",
+)
 @pytest.mark.parametrize(
     "youtube_uri, expected",
     [
-        (EXAMPLE_YOUTUBE_VIDEO, "XALBGkjkUPQXALBGkjkU.mp4"),
+        (EXAMPLE_YOUTUBE_VIDEO_EMBEDDED, "XALBGkjkUPQ.mp4"),
+        (EXAMPLE_YOUTUBE_VIDEO_PARAMETER, "XALBGkjkUPQ.mp4"),
+        (EXAMPLE_YOUTUBE_VIDEO_SHORT, "XALBGkjkUPQ.mp4"),
     ],
 )
 def test_youtube_downloader(
--- a/cdp_backend/tests/utils/test_file_utils.py
+++ b/cdp_backend/tests/utils/test_file_utils.py
@@ -1,9 +1,8 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-from __future__ import annotations
-
 import os
+import random
 import sys
 from pathlib import Path
 from typing import Optional
@@ -20,7 +19,11 @@ from cdp_backend.utils.file_utils import (
     resource_copy,
 )
 
-from ..conftest import EXAMPLE_VIDEO_FILENAME, EXAMPLE_VIDEO_HD_FILENAME
+from ..conftest import (
+    EXAMPLE_MKV_VIDEO_FILENAME,
+    EXAMPLE_VIDEO_FILENAME,
+    EXAMPLE_VIDEO_HD_FILENAME,
+)
 
 #############################################################################
 
@@ -164,7 +167,7 @@ def test_static_thumbnail_generator(
     "video_url, session_content_hash, num_frames, expected",
     [
         (EXAMPLE_VIDEO_FILENAME, "example2", 15, "example2-hover-thumbnail.gif"),
-        (EXAMPLE_VIDEO_HD_FILENAME, "example3", 9, "example3-hover-thumbnail.gif"),
+        (EXAMPLE_VIDEO_HD_FILENAME, "example3", 2, "example3-hover-thumbnail.gif"),
         pytest.param(
             "fake.mp4",
             "example",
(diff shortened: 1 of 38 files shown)


cdp_backend/utils/file_utils.py
```diff
@@ -126,6 +134,30 @@ def resource_copy(
         raise e
 
 
+def youtube_copy(youtube_url: str, output: str) -> str:
```
isaacna, line 137: We should consider `output` as `Optional[str]` and if there's no output name specified, hashing by the file contents. 

This would keep things similar to how we hash other files here: https://github.com/CouncilDataProject/cdp-backend/commit/873dddbcbfb6051c1c29d5cf9deab2eea4aa549a

@JacksonMaxfield thoughts?

cdp_backend/utils/file_utils.py
```diff
@@ -126,6 +134,30 @@ def resource_copy(
...
+        The url of the YouTube video to copy.
+    output: str
+        The location of the file to download.
+
+    Returns
+    _______
+    output: str
+        The location of the downloaded file.
+    """
+    if output[len(output) - 4 : len(output)] != ".mp4":
```
isaacna, line 153: I think it'd be cleaner to check the extension like this: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/event_gather_pipeline.py#L323-L327

cdp_backend/utils/file_utils.py
```diff
@@ -105,6 +106,13 @@ def resource_copy(
     log.info(f"Beginning resource copy from: {uri}")
     # Get file system
     try:
+        if uri.find("https://www.youtube.com/embed/") == 0:
```
isaacna, line 109: Can we expand this to run on youtube url's without `/embed` as well? The portland use case might only contain embedded videos, but we should try to make this function more flexible for future use cases

isaacna (changes requested):
Looks good overall! Just had a few comments about naming the file and accepting more url formats

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -224,3 +225,21 @@ def test_convert_video_to_mp4(
...
     assert file_utils.convert_video_to_mp4(filepath) == str(resources_dir / expected)
+
+
+@pytest.mark.parametrize(
+    "youtube_uri, expected",
+    [
+        (EXAMPLE_YOUTUBE_VIDEO, "XALBGkjkUPQXALBGkjkU.mp4"),
+    ],
+)
+def test_youtube_downloader(
```
JacksonMaxfield, line 236: Can you add this `skipif` to this test: https://github.com/CouncilDataProject/cdp-backend/blob/f95304853f01ce9d59db608dffd68a4099fbe539/cdp_backend/tests/database/test_validators.py#L85

Don't want to run this test if the person doesn't have internet available... (I code of the train a lot)

cdp_backend/utils/file_utils.py
JacksonMaxfield: Completely agree. I think there are multiple paths we want to check:

The base youtube.com and the youtube url shortener "youtu.be"

cdp_backend/utils/file_utils.py
JacksonMaxfield: I think my general comment is that I think this function should be folded into `resource_copy`? Maybe keep it as a separate function for now but _use_ the function in `resource_copy` so that that single function can be used to fetch all sorts of content.

JacksonMaxfield: Additionally, yes, if we could mirror the API of `resource_copy` that would be great. `output` -> renamed to `dst` and add an `overwrite` param as well.

Iirc, the `resource_copy` API matches the fsspec `.get` API so if they all match we can somewhat use them interchangably.

cdp_backend/utils/file_utils.py
JacksonMaxfield: Agree but I will say that if this is folded into / used directly in the `resource_copy` function then hopefully we don't need to have duplicate code.

cdp_backend/utils/file_utils.py
isaacna: > I think my general comment is that I think this function should be folded into resource_copy? Maybe keep it as a separate function for now but use the function in resource_copy so that that single function can be used to fetch all sorts of content.

Good point, forgot about `resource_copy`. I like the idea more of keeping the function separate but using it inside of `resource_copy`.

Shak2000: I changed the parameters' names to be identical to those in `resource_copy`. In my case, `dst` is guaranteed to exist since it is called from `resource_copy`. If we want to assume it's not guaranteed, then it might be better to transfer the code that makes up `youtube_copy` to `resource_copy` since I would need to duplicate the code if `dst` were empty.

cdp_backend/utils/file_utils.py
JacksonMaxfield: Hah. Wow, our bad @Shak2000, we both reviewed this and didn't realize it was in the `resource_copy` function already.

That said, I think my prior statement of "checking for all the various youtube link types" still stands.

cdp_backend/utils/file_utils.py
JacksonMaxfield: I _think_ I am okay with leaving `dst` as required then. When running this function on any youtube video, what is the produced file name?

Shak2000: If the video is not embedded, the file name would be the section of the YouTube video URL after `watch?v=`. For example, if the video URL is `https://www.youtube.com/watch?v=lheapd7bgLA`, the file name would be `lheapd7bgLA.mp4`.

If the video is embedded, the file name would be the section of the YouTube video URL after `embed/`. For example, if the video URL is `https://www.youtube.com/embed/JTW-jJzmPzQ`, the file name would be `JTW-jJzmPzQ`.

I chose this strategy since I know that YouTube always generates unique ID's for its videos.

I will soon update this pull request to accommodate these changes.

JacksonMaxfield: Perfect. That works for me.

cdp_backend/utils/file_utils.py
```diff
@@ -91,7 +91,10 @@ def resource_copy(
     # Ensure dst doesn't exist
     dst = Path(dst).resolve()
     if dst.is_dir():
-        dst = dst / uri.split("/")[-1]
+        if "=" in str(uri):
```
JacksonMaxfield, line 94: Nitpick: can we be even more specific here? does split by `v=` work?

And a comment that says "split by youtube video query parameter"

JacksonMaxfield (approved):
One minor nitpick but everything else looks perfect. Thanks for making the changes!

isaacna (approved):

cdp_backend/utils/file_utils.py
```diff
@@ -126,6 +135,37 @@ def resource_copy(
...
+    overwrite: bool
+        Boolean value indicating whether or not to overwrite a local video with
+        the same name if it already exists.
+
+    Returns
+    _______
+    dst: str
+        The location of the downloaded file.
+    """
+    dst = Path(str(dst) + ".mp4")
```
JacksonMaxfield, line 157: Was about to merge but one last look I think catches this.

This will always overwrite the supplied `dst` correct?

I think there simply needs to be an `if dst is not None` before this statement.

Additionally the docstring / parameter typing should be updated to `dst: Optional[Path] = None`

Shak2000: No, I assumed that `dst` is not empty since if `dst` is `None`, it would be a assigned a value, as seen [here](https://github.com/Shak2000/cdp-backend/blob/91fc71f7b5726b763715e385607436279177e32f/cdp_backend/utils/file_utils.py#L88). I can throw an exception if `dst` is `None` though. WDYT?

JacksonMaxfield: Ohhhh i see what you are doing. Yes. My mistake. Okay so then only nitpick would be instead of doing this `str(dst) + ".mp4"` string addition. `Path` has a function called `with_suffix` that is "safer" imo.

```python
Path(str(dst)).with_suffix(".mp4")
```

JacksonMaxfield: But yes, you are correct. My mistake on on the overwrite vs simply adding the mp4

Shak2000: I just pushed a new change based on your feedback.

PR #176, opened 2022-03-19
feature/https-or-http-func

### LINKED ISSUE #148: Add a function for replacing all http URIs with https URIs when possible
We already do this for session video URIs: https://github.com/CouncilDataProject/cdp-backend/blob/294254928cfc9ff1861d60f2208c91dd533b15e6/cdp_backend/pipeline/event_gather_pipeline.py#L337

However this should be the default standard whenever we check an HTTP like URI.

If given an http(s) uri, try the uri in https first, then try it in http, whichever works, use it.

liamphmurphy:
### Link to Relevant Issue

This pull request resolves #148 

### Description of Changes

This adds a func `try_url` in `string_utils.py` that will do a request to the https / http version of a given url, giving preference to https. If https is found first, it returns that url immediately and doesn't attempt to find the http version.

I'm passing in the optional resolve_func to allow for mock testing in unit tests, as I don't want to rely on external URL's for the tests to pass. The default resole_func is the already existing `resource_exists` func.


JacksonMaxfield (commented):
Will review tomorrow!!

liamphmurphy:
> Will review tomorrow!!

Feel free to take a look though it may not be in its final form, I forgot to comment that I wanted to get a PR open to confirm some test failures that I was seeing locally, looks like they are so I'll need to look into that.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -336,12 +338,14 @@ def convert_video_and_handle_host(
...
+    if not is_secure_uri(resource_uri):
         # Attempt to find secure version of resource and simply swap
         # otherwise we will have to host
-        if session.video_uri.startswith("http://"):
-            secure_uri = session.video_uri.replace("http://", "https://")
+        if resource_uri.startswith("http://"):
+            secure_uri = resource_uri.replace("http://", "https://")
             if resource_exists(secure_uri):
                 log.info(
                     f"Found secure version of {session.video_uri}, "
```
isaacna, line 351: This section seems a little redundant since we're already doing logic to check whether the url is `https` or `http` in `try_url`. I think we could tweak this to account for that

cdp_backend/utils/string_utils.py
```diff
@@ -147,3 +150,40 @@ def convert_gcs_json_url_to_gsutil_form(url: str) -> str:
...
+    resolve_func: func(url: str) -> bool
+        A function that takes in a str URL and determines whether it is reachable. Default is our "resource_exists" func
+
+    Returns
+    -------
+    resource_url: str
+        The url with the correct protocol based on where the resource exists. If does not exist, return empty str
+    """
+    if not url.startswith("http://") and not url.startswith("https://"):
+        raise ValueError("url must be a valid http or https url")
```
isaacna, line 172: I don't think we can say this is necessarily true in al cases. Certain url's pointing to GCP resources could start with `gs://`, and if so we handle those [differently](https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/validators.py#L120-L130)

So I wanna say we can delete this conditional statement

cdp_backend/utils/string_utils.py
```diff
@@ -147,3 +150,40 @@ def convert_gcs_json_url_to_gsutil_form(url: str) -> str:
...
+
+    if url.startswith("http://"):
+        url = url.replace("http://", "https://")
+
+    if resolve_func(url):
+        return url
+
+    url = url.replace("https://", "http://")
+    if resolve_func(url):
+        return url
```
isaacna, line 182: Instead of having two replace operations, could we make this something like: 
```python

    secure_url = url.replace("http://", "https://")

    if resolve_func(secure_url):
        return secure_url
    if resolve_func(url):
        return url

```

cdp_backend/utils/string_utils.py
```diff
@@ -147,3 +150,40 @@ def convert_gcs_json_url_to_gsutil_form(url: str) -> str:
...
+
+    # raise LookupError("the resource {} does not exist with either http / https".format(url))
+    return ""  # if we got this far, the resource does not exist
```
isaacna, line 185: I'm a little hesitant to return an empty string here instead of raising an error. We'd have to handle the empty string wherever this method is used, and if we miss that then it could lead to us accidentally processing data with an empty string instead of catching an error

cdp_backend/utils/string_utils.py
```diff
@@ -147,3 +150,40 @@ def convert_gcs_json_url_to_gsutil_form(url: str) -> str:
...
+        url = url.replace("http://", "https://")
+
+    if resolve_func(url):
+        return url
+
+    url = url.replace("https://", "http://")
+    if resolve_func(url):
+        return url
+
+    # raise LookupError("the resource {} does not exist with either http / https".format(url))
```
isaacna, line 184: This isn't necessarily true if the url is `gs://`

cdp_backend/tests/utils/test_string_utils.py
```diff
@@ -107,3 +107,22 @@ def test_clean_text(
 )
 def test_convert_gcs_json_url_to_gsutil_form(text: str, expected: str) -> None:
     assert string_utils.convert_gcs_json_url_to_gsutil_form(text) == expected
+
+
+@pytest.mark.parameterize(
+    "url, expected"[("https://exists", True), ("https://not-exists", False)],
```
isaacna, line 113: Think you missed a comma after `"url, expected"`

isaacna (changes requested):
Thanks for addressing this issue! I had a few comments, mostly regarding some edge cases where the url does not start with `http` or `https`. I think this could be improved by removing that assumption, and passing in the `resource_exists` callable that also has `google_credentials_file` as a kwarg so we can account for `gs://` cases

liamphmurphy:
> Thanks for addressing this issue! I had a few comments, mostly regarding some edge cases where the url does not start with `http` or `https`. I think this could be improved by removing that assumption, and passing in the `resource_exists` callable that also has `google_credentials_file` as a kwarg so we can account for `gs://` cases

Thanks for the feedback! Most of these comments should be addressed in some local changes I have which I'll push up soon. 

I tried to make `resource_exists` the default callable but was having import errors, but I shall try that again.

JacksonMaxfield:
> I tried to make `resource_exists` the default callable but was having import errors, but I shall try that again. On second look, it may be a circular dependency problem as `cdp_backend.database.validators` is importing from string_utils, and importing resource_exists would require pulling from `cdp_backend.database.validators`. I'm not the strongest with Python but I assume circular dependencies is considered a problem?

Yea circular dependencies are a no go in Python. I would be fine if you moved this function and associated tests from `string_utils` to `file_utils`... I don't think there is a circular dep there. If there is.... I guess just move it straight to `validators` for now?

liamphmurphy:
Made some changes, let me know what you all think. Validators made the most sense to me just given the name for putting this logic in, but will happily defer to you two and will change it to file_utils if that makes the most sense. 

Tests are passing locally but I'm seeing some weirdness with my setup (probably just missing python versions) so I don't 100% believe that yet.

JacksonMaxfield (approved):
This is great! Thanks for taking this one. I have two minor nitpicks but would be happy to merge now or after you answer one question!

1. One nitpick is on using `fstrings` instead of `string.format`
2. Can you lint the code real quick? Running `tox -e lint` should help fixup any of those issues. Related... GitHub is showing you deleted a test resource and added a test resource?

Question! This function is meant to help us archive "EventMinutesItemFile" and "MatterFile" data _better_.

If you take a quick look at the database schema: https://councildataproject.org/cdp-backend/database_schema.html

In the pipeline, we just ignore failing URI checks for matter file and such: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/event_gather_pipeline.py#L1444

and skip to the next task in the pipeline. It would be great if we used this function to try the URL before even creating and uploading the model. Basically, the pseudocode would be something like: `try_url` -> returned string -> create and upload model -> catch and log any random failure?? || `try_url` -> raised `LookupError` -> log
file details as unable to uplaod like we are currently doing

So, with that all said, want to include that work in this PR or in a second PR?

---

Also last note: how was the self-onboarding process? places we can improve?

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -337,24 +337,22 @@ def convert_video_and_handle_host(
...
-            if resource_exists(secure_uri):
+    elif not is_secure_uri(session.video_uri):
+        try:
+            resource_uri = try_url(session.video_uri)
+        except LookupError:
+            # The provided URI could still be like GCS or S3 URI, which works for download
+            # but not for streaming / hosting
+            cdp_will_host = True
+        else:
+            if is_secure_uri(resource_uri):
```
isaacna, line 348: Wouldn't this branch never hit? If `try_url` receives a `LookupError` then that would mean it couldn't access via both `https`  and `http`

cdp_backend/tests/database/test_validators.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+from multiprocessing.sharedctypes import Value
```
isaacna, line 4: Can you remove this unused import?

isaacna (changes requested):
Just had a question about the conditional logic + some nitpicks, and I think there were some extra files/imports that weren't being used. 

Also for PR's we use a [lint checker](https://github.com/CouncilDataProject/cdp-backend/runs/5618696766?check_suite_focus=true) which seems to be failing in this case. It's usually pretty descriptive about what it wants you to reformat, and the main checks we run are `isort`, `black`, `mypy`, and `flake`

cdp_backend/tests/utils/test_string_utils.py
```diff
@@ -107,3 +107,4 @@ def test_clean_text(
 )
 def test_convert_gcs_json_url_to_gsutil_form(text: str, expected: str) -> None:
     assert string_utils.convert_gcs_json_url_to_gsutil_form(text) == expected
+
```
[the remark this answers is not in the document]
liamphmurphy: The linter (`tox -e lint`) seems to be automatically adding that line back, unless I'm missing something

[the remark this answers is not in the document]
JacksonMaxfield: Yea! no worries. Trust the linter.

cdp_backend/pipeline/event_gather_pipeline.py
liamphmurphy: Are you talking about the `else` branch? My understanding is that runs if no exceptions occurred, and I just borrowed the logic for flagging that cdp will self host if the url is http, or setting the `hosted_video_media_url` to that url if it was secure. So the else branch, I think, would still need some delineation between handling http vs https URL's. But let me know if you're talking about something else

liamphmurphy:
> 

Thanks again for the feedback! I believe I addressed the two nitpicks you mentioned.

I'm happy to work on the archiving improvements as a part of this PR if this functionality not existing isn't blocking anything else. May take me a bit before I can fully address this next phase of the PR.

As for the onboarding, I think you all have an excellent onboarding process. My struggles are just not being that well versed with the Python ecosystem, mainly two things:
1. I'm coming from a very different (and sometimes opinionated) set of tools in the Go world. For example, I did the `"{}".format(str)` because that is how I'm used to doing it in Go with their (somewhat) equivalent `fmt.Sprintf("%s", str)`. 
2. I've never used tox before so I didn't have the linter setup to run on each save, which is how I have it done with my Go setup. So I've gotten very comfortable with Go's linter automatically removing unused imports among other things. To my eyes, the tox output from `make build` is pretty lengthy and verbose so I wasn't adequately spotting some of the errors it was notifying me about. Hopefully I'm a bit better keyed into that now.

I think the code base is excellent and I enjoyed working on this, so if you will have me on further things, I would be happy to receive any recommendations for a better Python dev setup.

cdp_backend/pipeline/event_gather_pipeline.py
isaacna: Gotcha I see, I misread the indentation bc the control flow was a little hard to read. This makes sense given the existing logic but I had a general question about this previously existing logic

@JacksonMaxfield is there a specific reason why we want CDP to host resources that arent secure? Is this a mixed content issue when storing http URL's? I saw this was added [here](https://github.com/CouncilDataProject/cdp-backend/pull/138) but wasn't explicitly mentioned

isaacna (approved):
Looks good to me, I had one question regarding mixed content for Jackson in a comment, but once that's resolved I think this can be merged in

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: Yea after a mental walkthrough of the control flow I think this makes sense. And yes, it's surprising to me that Chrome allows `http` videos to play from an `https` URL (i.e. our `https://councildataproject.org` links to a `http` video.

Firefox makes any resource request from `https` to `http` as unsafe and blocks the call

JacksonMaxfield: Basically, that is the reason. If the video isn't available under an `https` host, we need to host it because Firefox rejects the playback

JacksonMaxfield:
Going to merge this now simply because it is a nice unit of work to PR in. But yes please do work on the archiving improvements whenever you get a chance! Really really quick and neat work of this!

> I've never used tox before so I didn't have the linter setup to run on each save, which is how I have it done with my Go setup. So I've gotten very comfortable with Go's linter automatically removing unused imports among other things. To my eyes, the tox output from make build is pretty lengthy and verbose so I wasn't adequately spotting some of the errors it was notifying me about. Hopefully I'm a bit better keyed into that now.

Yea, we may want to change the CONTRIBUTING doc to mention that you can run individual versions of the build system with the environment specifier... `tox -e py38` will run python 3.8, `tox -e lint` will just run the linter etc.

As far as integration with IDE... yea, Python just doesn't have the build systems as Go or Rust do. Pretty good tho. VS Code can format on save and reorder imports and such on save but wont handle the lint and typing for you :/

PR #186, opened 2022-05-02
feature/add-resiliency-to-supporting-file-checking

### LINKED ISSUE #179: Try URL before creating / uploading Matterfile
### Feature Description

This feature (enhancement) would add a bit more resiliency to the creation of Matterfile's seen here: https://github.com/CouncilDataProject/cdp-backend/blob/2e472b92b5ef9cd49fc940bacfe1ed7e4f9cc496/cdp_backend/pipeline/event_gather_pipeline.py#L1444-L1452

Specifically, we want to check that the `supporting_file.uri` used in the creation of the Matterfile actually exists before creating / uploading the Matterfile.

### Use Case

Simply the use case here is to create better Matterfile data by ensuring that the URL's used exist.

### Solution

In the above linked code, before the call to `create_matter_file`, do the following:

1. Make a call to `try_url` (using the default resolve_func) to see if the URL exists.
2. IF exists, proceed with the creation / uploading of the Matterfile.
3. IF not exists, add `LookupError` as an exception to watch for in that loop, as `try_url` will raise that exception on errors.

### Alternatives

Let me know if any other options are better.

liamphmurphy:
### Link to Relevant Issue

This pull request resolves #179 

### Description of Changes

This PR adds a try_url call for the supporting_file.uri used before the creation / upload of a matter file (the try is done before these actions occur). If a LookupError is thrown, then perform a usual log.

### DIFF
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -1458,7 +1458,6 @@ def store_event_processing_results(
                         except (
                             FieldValidationFailed,
                             ConnectionError,
-                            LookupError,
                         ) as e:
                             log.error(
                                 f"MatterFile ('{supporting_file.uri}') "
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -1442,6 +1442,7 @@ def store_event_processing_results(
                             f"SupportingFile ('{supporting_file.uri}') "
                             f"uri does not exist. Skipping. Error: {e}"
                         )
+                        continue
 
                     # Archive as matter file
                     if event_minutes_item.matter is not None:
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -1434,13 +1434,18 @@ def store_event_processing_results(
             # Add supporting files for matter and event minutes item
             if event_minutes_item.supporting_files is not None:
                 for supporting_file in event_minutes_item.supporting_files:
+                    try:
+                        file_uri = try_url(supporting_file.uri)
+                        supporting_file.uri = file_uri
+                    except LookupError as e:
+                        log.error(
+                            f"SupportingFile ('{supporting_file.uri}') "
+                            f"uri does not exist. Skipping. Error: {e}"
+                        )
 
                     # Archive as matter file
                     if event_minutes_item.matter is not None:
                         try:
-                            file_uri = try_url(supporting_file.uri)
-                            supporting_file.uri = file_uri
-
                             matter_file_db_model = db_functions.create_matter_file(
                                 matter_ref=matter_db_model,
                                 supporting_file=supporting_file,
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -1438,6 +1438,9 @@ def store_event_processing_results(
                     # Archive as matter file
                     if event_minutes_item.matter is not None:
                         try:
+                            file_uri = try_url(supporting_file.uri)
+                            supporting_file.uri = file_uri
+
                             matter_file_db_model = db_functions.create_matter_file(
                                 matter_ref=matter_db_model,
                                 supporting_file=supporting_file,
@@ -1447,7 +1450,11 @@ def store_event_processing_results(
                                 db_model=matter_file_db_model,
                                 credentials_file=credentials_file,
                             )
-                        except (FieldValidationFailed, ConnectionError) as e:
+                        except (
+                            FieldValidationFailed,
+                            ConnectionError,
+                            LookupError,
+                        ) as e:
                             log.error(
                                 f"MatterFile ('{supporting_file.uri}') "
                                 f"could not be archived. Skipping. Error: {e}"
--- a/cdp_backend/tests/pipeline/test_event_gather_pipeline.py
+++ b/cdp_backend/tests/pipeline/test_event_gather_pipeline.py
@@ -538,9 +538,7 @@ def test_store_event_processing_results(
     # externally either.
     mock_resource_copy.return_value = "doesnt-matter.ext"
 
-    with mock.patch(
-        "cdp_backend.database.validators.resource_exists"
-    ) as mock_resource_exists:
+    with mock.patch(f"{PIPELINE_PATH}.resource_exists") as mock_resource_exists:
         if fail_try_url:
             mock_resource_exists.side_effect = LookupError()
         else:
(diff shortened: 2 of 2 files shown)


cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -1438,6 +1438,9 @@ def store_event_processing_results(
                     # Archive as matter file
                     if event_minutes_item.matter is not None:
                         try:
+                            file_uri = try_url(supporting_file.uri)
```
JacksonMaxfield, line 1441: Can you pull this out a level above so that you can `try_url` once and use the "correct" url for both the matter file and the minutes item file: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/event_gather_pipeline.py#L1457

JacksonMaxfield (commented):
Love the test addition. Minor comment on where to place this is the pipeline as a whole but it _should_ be an easy fix I think.

cdp_backend/pipeline/event_gather_pipeline.py
liamphmurphy: Ah good catch, was tunnel-visioned on the matter file

cdp_backend/tests/pipeline/test_event_gather_pipeline.py
```diff
@@ -45,6 +45,7 @@
 # great system stdlib :upsidedownface:
 
 PIPELINE_PATH = "cdp_backend.pipeline.event_gather_pipeline"
+DATABASE_PATH = "cdp_backend.database"
```
dphoria, line 48: Is this used anywhere?

liamphmurphy: 👍 artifact from when I was having issues with the mock, ty

dphoria (approved):
Thank you Liam. Looks good to me, pending CI test results.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -1434,6 +1434,14 @@ def store_event_processing_results(
...
             if event_minutes_item.supporting_files is not None:
                 for supporting_file in event_minutes_item.supporting_files:
+                    try:
+                        file_uri = try_url(supporting_file.uri)
+                        supporting_file.uri = file_uri
+                    except LookupError as e:
+                        log.error(
+                            f"SupportingFile ('{supporting_file.uri}') "
+                            f"uri does not exist. Skipping. Error: {e}"
+                        )
```
JacksonMaxfield, line 1444: I have one extra change. This try-except block will run but then regardless of error or not the follow sections will run. In the `except` block, I think we should add a `continue`?

https://www.tutorialspoint.com/python/python_continue_statement.htm

I have never used a continue but this seems like the perfect time to??

liamphmurphy: Good point, the control statements would have stopped further execution on an exception before. 

But agreed, `continue` is the right case here (I assume we don't want to fail the processing of all supporting files just because one fails)

dphoria: Nice catch again. I concur.

JacksonMaxfield: > (I assume we don't want to fail the processing of all supporting files just because one fails)

Correct assumption!

liamphmurphy: codecov is saying the except block is never reached, I would have thought it would with the `fail_try_url == True` cases...

cdp_backend/tests/pipeline/test_event_gather_pipeline.py
```diff
@@ -523,12 +531,21 @@ def test_store_event_processing_results(
...
     # All of the resource copies relate to image file uploads / archival.
     # But we aren't actually uploading so just make sure that we aren't downloading
     # externally either.
     mock_resource_copy.return_value = "doesnt-matter.ext"
 
+    with mock.patch(
+        "cdp_backend.database.validators.resource_exists"
+    ) as mock_resource_exists:
+        if fail_try_url:
+            mock_resource_exists.side_effect = LookupError
```
dphoria, line 545: Should this be `LookupError()`?

cdp_backend/pipeline/event_gather_pipeline.py
JacksonMaxfield: Hmmm technically you are testing `resource_exists` do you want to explicitely test `try_url`?

liamphmurphy: I didn't want to get into testing `try_url` directly since it could return a range of values depending on the test, I thought `resource_exists` raising LookupError in the mock would pass it up through `try_url` back to the caller, but I could be wrong.

JacksonMaxfield: It _should_ but it may be a weird pytest module pathing bug. I will look in a second.

liamphmurphy: Same result if setting the mock to return `False`, try_url will raise the error if `resource_exists` returns false. Strange.

JacksonMaxfield: If the `LookupError()` change doesn't work, I think the solution will be to replace the mock.patch string from: `"cdp_backend.database.validators.resource_exists"` to `f"{PIPELINE_PATH}.resource_exists"`

This is the most annoying thing about pytest. It mocks the loaded module, not the global module path.

liamphmurphy: Neither changes seem to be working based on my local coverage report, will push it up just to make sure, but if not I can double back to this or we can say it's not worth the complexity... I would trust the try / except logic to work correctly here.

JacksonMaxfield: I am going to leave some time for myself later in the week to figure this out. No worries though. Thanks for handling this!!!! :heart:

isaacna (approved):
Looks good to me in general! I agree with using `continue` in the lookup failure use case. I think regarding the test case not hitting the exception block, I think it's fine mocking `try_url` directly rather than `resource_exists`. We're testing the overall pipeline logic and can just mock `try_url` directly. 

I think if the use of `resource_exists` is functioning oddly when passing it to `try_url`, we can handle testing that logic specifically for that method separately in `test_validators.py`. My guess is that it's some quirk with mock or how the pathing works, which isn't a big deal to me. We can mock `try_url` so that we're abstracting the inner workings of that method from the pipeline tests.

isaacna:
If you really want to see if this is a bug in logic vs. an issue with mock, you can do a manual run of the pipeline and set the supporting file to something invalid and see what's happening

JacksonMaxfield:
Closing in favor of #187

PR #188, opened 2022-05-22
feature/support-processing-m3u8-video-playlist-uris

JacksonMaxfield:
### Link to Relevant Issue

https://github.com/CouncilDataProject/cdp-scrapers/pull/96#issuecomment-1114325009

### Description of Changes

Allows processing of m3u8 URIs.

What will happen is during `resource_copy` we will download and convert the m3u8 playlist to an mp4. Then during determining the video host, we check the original session video uri and see if it ends with `".m3u8"` and if so, we host the already converted and saved locally m3u8 file.

### DIFF
--- a/cdp_backend/utils/file_utils.py
+++ b/cdp_backend/utils/file_utils.py
@@ -7,6 +7,7 @@ import random
 from hashlib import sha256
 from pathlib import Path
 from typing import Optional, Tuple, Union
+from uuid import uuid4
 
 import fsspec
 import requests
@@ -116,12 +117,11 @@ def resource_copy(
         if uri.endswith(".m3u8"):
             import m3u8_To_MP4
 
-            # We add a hash of the uri to the front of the filename because m3u8 files
+            # We add a uuid4 to the front of the filename because m3u8 files
             # are usually simply called playlist.m3u8 -- the result will be
-            # f"{hash}-{name}"
-            uri_hash = sha256(uri.encode("utf-8")).hexdigest()
+            # f"{uuid}-{name}"
             mp4_name = dst.with_suffix(".mp4").name
-            save_name = f"{uri_hash}-{mp4_name}"
+            save_name = f"{uuid4()}-{mp4_name}"
 
             # Reset dst
             dst = dst.parent / save_name
--- a/cdp_backend/utils/file_utils.py
+++ b/cdp_backend/utils/file_utils.py
@@ -7,6 +7,7 @@ import random
 from hashlib import sha256
 from pathlib import Path
 from typing import Optional, Tuple, Union
+from uuid import uuid4
 
 import fsspec
 import requests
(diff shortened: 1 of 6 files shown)


dphoria (approved):
Appreciate!

JacksonMaxfield:
I have two actions running on my dev infrastructure to test:

1. testing the manual / custom process event: https://github.com/JacksonMaxfield/cdp-dev/runs/6545676305?check_suite_focus=true
2. testing the normal event gather: https://github.com/JacksonMaxfield/cdp-dev/runs/6545697019?check_suite_focus=true

if both succeed I will link to their page on my dev infra website and then likely merge after approvals! 🎉

cdp_backend/utils/file_utils.py
```diff
@@ -113,6 +114,26 @@ def resource_copy(
...
+
+            # Reset dst
+            dst = dst.parent / save_name
+
+            # Download and convert
+            m3u8_To_MP4.download(
+                uri,
+                mp4_file_dir=dst.parent,
+                mp4_file_name=save_name,
+            )
```
tohuynh, line 134: This downloads to an mp4 file in local storage - could that be a problem on GitHub actions with limited storage?

JacksonMaxfield: We store the video on the disk regardless of m3u8 or native mp4. This slightly duplicates storage because it stores chunks and the converted but 🤷 we have never run into problems. Even when processing 8 events at a time

tohuynh: I think the limit is [500mb](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes). Are the temp files deleted after they have been hosted, that would help?

JacksonMaxfield: 500Mb wouldn't even be possible to store a docker image on.

I think it's 14GB: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

tohuynh: The 500MB must be for logs and artifact files then 😕

JacksonMaxfield: Now that is likely

JacksonMaxfield:
Both the normal event gather and the manual process special event succeeded

https://jacksonmaxfield.github.io/cdp-dev/#/events/0913ecd21d43
https://jacksonmaxfield.github.io/cdp-dev/#/events/99f4c2a99447

@tohuynh you happy with this PR if so I think I can merge. 🎉

tohuynh:
> Both the normal event gather and the manual process special event succeeded
> 
> https://jacksonmaxfield.github.io/cdp-dev/#/events/0913ecd21d43 https://jacksonmaxfield.github.io/cdp-dev/#/events/99f4c2a99447
> 
> @tohuynh you happy with this PR if so I think I can merge. 🎉

yes

PR #189, opened 2022-06-02
feature/reduce-and-fan-ngram-index

JacksonMaxfield:
### Description of Changes

This makes our indexing pipeline two different functions / bin scripts.

The first is to generate a lot of small index parquet chunks. The second then uploads a single chunk.

This is intended to be used with GitHub Actions where the first script runs, uploads all the index chunks to artifact files, then spawns a new GitHub action runner for each chunk uploaded to then upload that single chunk.

Gather -> Process -> Store -> Fan -> Upload


JacksonMaxfield:
This is ready for review! There is a related cookiecutter-cdp-deployment PR here: https://github.com/CouncilDataProject/cookiecutter-cdp-deployment/pull/108

You can see this pipeline in action here: https://github.com/JacksonMaxfield/cdp-dev/actions/runs/2491590958

cdp_backend/pipeline/process_event_index_chunk_pipeline.py
```diff
@@ -0,0 +1,157 @@
...
+    Split the large n_grams dataframe into multiple lists of IndexedEventGram models
+    for batched, mapped, upload.
+    """
+    # Read index chunk
+    n_grams_df = pd.read_parquet(chunk_path)
+
+    # Split single large dataframe into many dataframes
+    chunk_size = 500
+    n_grams_dfs = [
+        n_grams_df[i : i + chunk_size]
```
tohuynh, line 57: Nit: I don't know how you feel about double list comprehension, but couldn't you create the inner `event_gram_chunk` with another list comprehension. List comprehension is faster than appending from what I remember.

Or create one list of almost-complete-grams with list comprehension and then chunk it with list comprehension (to avoid the double for loop and appending)?

cdp_backend/pipeline/process_event_index_chunk_pipeline.py
```diff
@@ -0,0 +1,157 @@
...
+) -> List[List[AlmostCompleteIndexedEventGram]]:
+    """
+    Split the large n_grams dataframe into multiple lists of IndexedEventGram models
+    for batched, mapped, upload.
+    """
+    # Read index chunk
+    n_grams_df = pd.read_parquet(chunk_path)
+
+    # Split single large dataframe into many dataframes
+    chunk_size = 500
```
tohuynh, line 55: Could be a good number to parameterize? (upload batch size?)

cdp_backend/pipeline/generate_event_index_pipeline.py
```diff
@@ -343,99 +346,52 @@ def compute_tfidf(
...
-            credentials_file=credentials_file,
-            batch=batch,
-        )
+    # Create storage dir
+    storage_dir.mkdir(parents=True)
 
-    # Commit
-    batch.commit()
+    # Split single large dataframe into many dataframes
+    chunk_size = 50_000
```
tohuynh, line 372: Could be a number to parametrize and avoid magic numbers in the code. Maybe something along the line of the file size of the chunk?

tohuynh (commented):
Nice work!

cdp_backend/file_store/functions.py
```diff
@@ -126,6 +126,17 @@ def upload_file(
         return save_url
 
 
+def download_file(
```
isaacna, line 129: Nitpick but should we name this something like `gcs_download_file`? Since we do have [`resource_copy`](https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/utils/file_utils.py#L64) which handles generic file downloads (and also GCS downloads, but not with explicit use of the google credentials

isaacna (approved):
Looks good to me! This is a really cool way of optimizing the index pipeline!

cdp_backend/pipeline/process_event_index_chunk_pipeline.py
JacksonMaxfield: This is a single list comprehension just with a dataframe index access getitem. That said, I will think about how to clean this up.

cdp_backend/file_store/functions.py
JacksonMaxfield: Ehhhhh. We have a function called upload_file in this module so this is just the mirror of that. It is also "scoped" to `file_store.functions` so I am going to leave it as is.

cdp_backend/pipeline/process_event_index_chunk_pipeline.py
JacksonMaxfield: Yea this is really just a getitem call within a list comprehension. To expand it to a full for loop would add a lot of redundent code imo. Leaving as is.

PR #193, opened 2022-06-13
bugfix/fix-newline-webvtt-sentence-segmentation

### LINKED ISSUE #190: WebVTT parsing fails for Boston
### Describe the Bug

WebVTT parsing isn't the greatest for Boston:
https://councildataproject.org/boston/#/events/3c7f0df0ecd4

### Expected Behavior

_What did you expect to happen instead?_

WebVTT parsing should be the same level or better than Seattle

isaacna:
### Link to Relevant Issue

This pull request resolves #190 

### Description of Changes

Realized that the reason why Boston's parsing is so weird is because the `.vtt` file has several `\n` characters. It turns out that for caption segments that have `\n' in them, the regex pattern for finding the end of the sentence fails. I believe it's for a reason related to [this](https://emacs.stackexchange.com/questions/9548/what-is-the-regex-to-match-a-newline-character).

Fixed the issue by replacing all `\n' characters with spaces. Testing this out on the same vtt file resulted in bumping up the number of sentences from 37 to 195. 

Here's an example of the first few sentences which are split with more granularity after the fix. ([Original](https://boston.granicus.com//videos/87/captions.vtt)):
```
uh uh, i'm the city councilor for district five and i'm the vice president of the boston city council.

Viewers can watch the council meeting live on youtube by visiting boston city council tv. i would like to ask my colleagues and those in the audience to please silence their phones and electronic devices.

Thank you. please also be respectful and do not disrupt the meeting while you are here.

If you disruptive you will be asked to leave and if you fail to comply you will be escorted out. please also note the fact that according city council rules there are no signs allowed in the chamber.

Mr. clark, would you please call the roll to ascertain the presence of a quorum? councilor arroyo present councilor baker, councilor book councilor breadon councilor colletta councilor for ananda sanderson, councilor flaherty councilor flynn councilor of council who is general councilor.

Let me here councilor murphy and councilor arcore. thank you mr clark.

I have been informed by the clerk that a quorum is present.

This week's clergy is father bob carr of st. anthony parish in austin invited by councilor breadon. councilor breadon, would you like to come up to the podium, introduce our clergy for today ?
```


JacksonMaxfield (commented):
Hah wow. Only thing I am going to request is to add a test to webvtt parser tests with a boston generated VTT file.

isaacna:
Tested this on my local setup and seems that the tests fail intermittently, and I'm not sure why. Will look into this in a bit

isaacna:
For context into why the test case was previously failing, `webvtt` uses truecasing for certain words that may be capitalized differently ("rideshare "-> "RIDESHARE"). This is non-deterministic however so was causing inconsistent test results. 

Reduced the file size and only included words that didn't trigger this issue

JacksonMaxfield (approved):
Looks good to me! Thanks!

Issue #131, opened 2021-11-05
Speaker classification  


isaacna:
### Feature Description

Adding speaker classification to CDP transcripts. This could be through a script/class that retroactively attaches the speaker name to a transcript that already has [speaker diarization enabled](https://cloud.google.com/speech-to-text/docs/multiple-voices). Prodigy can be used for annotating the training data.

### Use Case

With speaker classification we can provide transcripts annotated with the speaker. This can be used in many ways such as through a script or github action

### Solution

Very high level idea would be to:
- use GCP's built-in speaker diarization to separate the speakers. We could also use something like Prodigy to annotate the data, but I believe they have their own diarization/transcription models as well. 
- Figure out how to add the classified speaker names to the diarized transcript. I'm not sure if GCP allows you to provide any training data, but from what I could tell they only separate the speakers, but the models don't take in training data to label them.

isaacna:
@JacksonMaxfield feel free to add any thoughts/ideas!

isaacna:
I think the most difficult part will be attaching identities to transcripts that are already speaker differentiated

JacksonMaxfield:
> Use GCP's built-in speaker diarization to separate the speakers. We could also create our own audio classification model. We could also use something like Prodigy to annotate the data, but I believe they have their own diarization/transcription models as well.

We can't use speaker diarization. It has a limit of 8 speakers because it uses some clustering under the hood or they are just arbitrarily imposing some limit.

The best method in my mind would be to fine tune a speech classification model with our own labels (gonzalez, sawant, mosqueda, etc).

https://huggingface.co/superb/wav2vec2-base-superb-sid

Here is a decent notebook on fine tuning and using the hugging face / transformers API for fine tuning an existing transformer: https://colab.research.google.com/github/huggingface/notebooks/blob/master/examples/audio_classification.ipynb

The idea in my head is:

1. collect ~20 clips of each council member speaking
2. to create a dataset, create random slices of each clip that are say 5 seconds in length each
3. This should blow up your 20 clips * council members * 5 second random spans

(I am arbitrarely choosing 5 seconds here because I don't know how long of an audio clip these pretrained audio classifiers allow in / how much memory is needed to train -- if we can fit a 30 second audio clip in then just do that.... and replace any reference I make to 5 seconds with 30 seconds -- find the audio clip size that fits in memory and performs well)

[...]

PR #197, opened 2022-06-29
feature/speaker-classifier-apply-function

JacksonMaxfield:
### Link to Relevant Issue

WIP #131 

### Description of Changes

Adds the function to annotate a transcript with a trained speakerbox model!

I am already using this and applying the highest accuracy model for Seattle to all of 2021-01-01 to 2022-01-01 data:

* Application function: https://github.com/JacksonMaxfield/phd-infrastructures/blob/main/speakerbox/manager.py#L532
* Log for just 2021-01-01 to 2021-02-01: https://github.com/JacksonMaxfield/phd-infrastructures/runs/7122165087?check_suite_focus=true


cdp_backend/annotation/speaker_labels.py
```diff
@@ -0,0 +1,201 @@
...
+DEFAULT_MODEL = "trained-speakerbox"
+
+###############################################################################
+
+
+def annotate(
+    transcript: Union[str, Path, Transcript],
+    audio: Union[str, Path, AudioSegment],
+    model: str = DEFAULT_MODEL,
+    min_intra_sentence_chunk_duration: float = 0.5,
```
isaacna, line 31: Do you think 0.5 is enough for a minimum? For certain things like roundtable one-word answers like 'yes/no' I could see it taking slightly less than half a second

cdp_backend/annotation/speaker_labels.py
```diff
@@ -0,0 +1,201 @@
...
+    max_intra_sentence_chunk_duration: float = 2.0,
+    min_sentence_mean_confidence: float = 0.985,
```
isaacna, line 33: Is this based on the confidence you've been seeing with the existing model you trained?

cdp_backend/annotation/speaker_labels.py
```diff
@@ -0,0 +1,201 @@
...
+    Transcript
+        The annotated transcript. Note this updates in place, this does not return a new
+        Transcript object.
+
+    Notes
+    -----
+    A plain text walkthrough of this function is as follows:
+
+    For each sentence in the provided transcript, the matching audio portion is
+    retrived, for example if the sentence start time is 12.05 seconds and end time is
```
isaacna, line 74: Nitpick `retrived -> retrieved`

setup.py
```diff
@@ -27,13 +27,18 @@
...
+    "pandas~=1.2",
     "prefect~=1.2",
     "pyarrow~=8.0",
+    "pydub~=0.25.1",
     "rapidfuzz~=2.0",
     "spacy~=3.0",
-    "truecase==0.0.14",
-    "webvtt-py==0.4.6",
+    "torch~=1.10",
+    "tqdm>=4.62",
```
isaacna, line 38: Is there a specific reason for `>=` instead of `~=`?

isaacna (approved):
Looks good to me! Just had a few questions about the some of the default values we picked for sentence length and confidence

cdp_backend/annotation/speaker_labels.py
```diff
@@ -0,0 +1,201 @@
...
+        Default: "trained-speakerbox"
+    min_intra_sentence_chunk_duration: float
+        The minimum duration of a sentence to annotate. Anything less than this
+        will be ignored from annotation.
+        Default: 0.5 seconds
+    max_intra_sentence_chunk_duration: float
+        The maximum duration for a sentences audio to split to. This should match
+        whatever was used during model training
+        (i.e. trained on 2 second audio chunks, apply on 2 second audio chunks)
+        Default: 2 seconds
```
tohuynh, line 56: What's the reasoning behind this?

cdp_backend/annotation/speaker_labels.py
```diff
@@ -0,0 +1,201 @@
...
+            loaded_transcript = Transcript.from_json(open_f.read())  # type: ignore
+    else:
+        loaded_transcript = transcript
+
+    # Load audio
+    if isinstance(audio, (str, Path)):
+        audio = AudioSegment.from_file(audio)
+
+    # Load model
+    classifier = pipeline("audio-classification", model=model)
```
tohuynh, line 111: How long does it take to load a model? Is one loading of the model per transcript efficient or can one loading of the model be passed around for a number of transcripts?

tohuynh (commented):
`annotate` looks good to me. Just have a couple of questions.

cdp_backend/annotation/speaker_labels.py
JacksonMaxfield: I think this is one of those things of... "is it valuable to the downstream analysis"

anything less than 0.5 seconds to me seems non-valuable to tag.

Though I am trying too find a better justification for this. Part of the reason is that I only trained on data that ranged from 0.5 - 2 seconds. And that was because anything less than 0.5 seconds worried me for "how much data to predict with" the smaller the clip the less information to use.

cdp_backend/annotation/speaker_labels.py
JacksonMaxfield: Hah. Great question and one that I am struggling with. Right now, this is 0.985 is based off: "i tried a bunch of different thresholds and this one seemed good" but I would love to know if there is a formula from confidence -> p-value??

cdp_backend/annotation/speaker_labels.py
JacksonMaxfield: Originally couldn't fit more than 2 seconds of audio into GPU during training.

cdp_backend/annotation/speaker_labels.py
JacksonMaxfield: It's fast! But yes I think it would be fair to say "path or preloaded model".

setup.py
JacksonMaxfield: Nope. Will make it ~=

PR #199, opened 2022-07-12
feature/vimeo-downloader

### LINKED ISSUE #191: Process vimeo video URLs
### Feature Description

Download and process videos hosted by vimeo.

### Use Case

Certain munis store videos and captions on Vimeo (chicago: https://vimeo.com/showcase/citycouncil)

### Solution

Likely a solution similar to the implementation of m3u8 processing: https://github.com/CouncilDataProject/cdp-backend/pull/188

There is a vimeo downloader python library: https://github.com/yashrathi-git/vimeo_downloader so we can use that for now but we should also make a PR to that library to add functionality to download the captions as well.

### Alternatives

Shak2000:
### Link to Relevant Issue

This pull request resolves #191 

### Description of Changes

Added a Vimeo Downloader. [Note: Does not work for every video on Vimeo](https://github.com/CouncilDataProject/cdp-backend/issues/191#issuecomment-1179828646). Still looking for a solution for that


cdp_backend/utils/file_utils.py
```diff
@@ -199,6 +201,40 @@ def youtube_copy(uri: str, dst: Path, overwrite: bool = False) -> str:
...
+    from vimeo_downloader import Vimeo
+
+    # Ensure dest isn't a file
+    if dst.is_file() and not overwrite:
+        raise FileExistsError(dst)
+
+    v = Vimeo(uri)
+
+    if len(v.streams) == 0:
+        raise Exception("File {} contains no downloadable streams", uri)
```
evamaxfield, line 232: Can you raise a `ValueError` instead of a base Exception.

cdp_backend/utils/file_utils.py
```diff
@@ -199,6 +201,40 @@ def youtube_copy(uri: str, dst: Path, overwrite: bool = False) -> str:
...
+    best_stream = v.streams[-1]
+    best_stream.download(download_directory=str(dst), filename=dst.name + ".mp4")
+    return str(dst) + "/" + dst.name + ".mp4"
```
evamaxfield, line 235: nitpick: prefer f-strings

`f"{dst}/{dst.name}.mp4"`

evamaxfield (commented):
Two very minor changes but otherwise looks good to me!!

Thanks @Shak2000

cdp_backend/utils/file_utils.py
evamaxfield: Bump this. Can you change this exception to a `ValueError`

setup.py
```diff
@@ -35,6 +35,7 @@
     "truecase==0.0.14",
     "webvtt-py==0.4.6",
     "yt-dlp>=2022.2.4",
+    "vimeo_downloader>=0.4.1"
```
evamaxfield, line 38: Could you change this to `~=0.4`

The youtube dl is using `>=` because it uses a different versioning scheme.

evamaxfield (commented):
Two minor changes then good to go!

cdp_backend/utils/file_utils.py
```diff
@@ -199,6 +201,40 @@ def youtube_copy(uri: str, dst: Path, overwrite: bool = False) -> str:
...
+    if dst.is_file() and not overwrite:
+        raise FileExistsError(dst)
+
+    v = Vimeo(uri)
+
+    if len(v.streams) == 0:
+        raise ValueError("File {} contains no downloadable streams", uri)
+    best_stream = v.streams[-1]
+    best_stream.download(download_directory=str(dst), filename=dst.name + ".mp4")
+    return f"{dst}/{dst.name}.mp4"
```
evamaxfield, line 235: My bad on requesting a bad change! I think you were correct in doing a path "division" the first time. This is what I would recommend. I think this one should work.

```python
return dst / f"{dst.name}.mp4"
```

evamaxfield (approved):
Assuming the tests pass, will merge! Thanks @Shak2000 !!

PR #204, opened 2022-08-20
feature/basic-gcp-function-and-clipping-gcp-function

evamaxfield:
~!!! No need to actually review this code, I just figured this would be a good place to ask questions !!!

---

This PR contains a very basic gcloud function that, when deployed, will respond to requests with a `name` parameter and say hello along with the shipped cdp-backend library version and the caller's IP.

CI has been updated to both lint, format, and type-check the function, and the infrastructure management has been updated to deploy the function and enable all the required services.

This won't get merged but it is easily enough of a proof of concept to show how we can easily do the video / audio clipping feature!

Which means that the tasks for the day of the hackathon are:
* write the library function to handle:
  * downloading a video file
  * clipping the video file
  * storing the clip with a "delete after storage time" setting
  * and returning the public download url
* add to the frontend:
  * button to show clipping model for params
  * take the params and make the request
  * wait for response
  * show the download url somehow

This was so easy to setup I am excited about other possibilities for functions to ship.

Only concern right now is that I am shipping this with "no auth" and I think that's _fine_, all of our stuff is no auth, but I want rate limiting potentially??

To make CI work I had to make a bunch of other small admin changes.

cc @Shak2000 (can't tag you as reviewer for some reason...)~

THIS IS THE CLIPPING FUNCTION ON THE BACKEND


.pre-commit-config.yaml
```diff
@@ -1,4 +1,3 @@
-files: cdp_backend
```
evamaxfield, line 1: Removed so that pre-commit checks the functions dir as well.

cdp_backend/infrastructure/Justfile
```diff
@@ -83,6 +83,11 @@ deploy project cookiecutter_yaml=default_cookiecutter_yaml:
 	store_cdp_metadata_document {{cookiecutter_yaml}}
 	gsutil cors set cors.json gs://{{project}}.appspot.com/
 	gsutil defacl ch -u AllUsers:R gs://{{project}}.appspot.com/
+	cd {{justfile_directory()}}/../gcloud_functions/hello-world/ && \
```
evamaxfield, line 86: This will be an interesting one. This is how it is laid out _right now in the dev infrastructure dir_ but come actual prod deployment time, we may just take a {functions_dir} path. Or actually, we should probably ship all of these functions with `cdp-backend` and unpack them with `get_cdp_infrastructure_stack`.

I leave that for a me task on the day of.

gcloud_functions/hello-world/.gcloudignore
```diff
@@ -0,0 +1 @@
+test_*.py
```
evamaxfield, line 1: Don't ship the test code.

gcloud_functions/hello-world/main.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    request_json = request.get_json(silent=True)
+    request_args = request.args
+
+    if request_json and "name" in request_json:
+        name = request_json["name"]
+    elif request_args and "name" in request_args:
+        name = request_args["name"]
+    else:
+        name = "World"
```
evamaxfield, line 31: Pulled from: https://cloud.google.com/functions/docs/create-deploy-http-python

Allows JSON args or query params

gcloud_functions/hello-world/main.py
```diff
@@ -0,0 +1,35 @@
...
+    return (
+        f"Hello {escape(name)} -- using CDP version: {__version__}! "
+        f"Caller: {request.remote_addr}"
```
evamaxfield, line 34: remote_addr is caller IP. Just for testing stuff.

PR #205, opened 2022-08-23
feature/support-undownloadable-vimeo-links

### LINKED ISSUE #191: Process vimeo video URLs

sagarrat7:
### Link to Relevant Issue

This pull request resolves #191

### Description of Changes

Use `Vimeo.from_video_id()` instead of `Vimeo()` when downloading from Vimeo. This supports downloading from URLs with the form: https://vimeo.com/showcase/6277394/video/722690793, as well as the previously supported form: https://vimeo.com/686512896.


cdp_backend/utils/file_utils.py
```diff
@@ -226,7 +227,11 @@ def vimeo_copy(uri: str, dst: Path, overwrite: bool = False) -> str:
     if dst.is_file() and not overwrite:
         raise FileExistsError(dst)
 
-    v = Vimeo(uri)
+    # 9 is the length of the video id; no other
+    # string of numbers in the URL has length 9 (thankfully)
+    match = re.findall(r"\d{9}", uri)
+    vid = match[0]
```
evamaxfield, line 233: Only comment is I think this may fail if the id wasn't found.

Mind making a catch here:

```python
if len(match) == 0:
   raise ValueError(f"Could not extract video id from uri: '{uri}'")
   
vid = match[0]
```

evamaxfield (approved):
Wow really nice work and quick find!

Only have one comment which is to make sure a result was found so that we can have a better error message in the case we didn't find anything.

PR #206, opened 2022-08-27
bugfix/webvtt-parsing-for-oakland

### LINKED ISSUE #203: Parse closed caption files for Oakland better
I fixed the cookiecutter / infra build and deployed Oakland!

There are still minor problems with permissions / CORS settings so ignore the broken video (I will fix that tomorrow)

That all said, Oakland has closed caption files which we parse and convert into our transcript format like we do for Seattle and Boston, once again, there is some minor differences from boston and seattle that I would love to be able to fix.

example of oakland event: https://councildataproject.org/oakland/#/events/e0912a619d69

the transcript includes a bunch of `">> "` markings where there are new speakers. To me these should be filtered out properly.

Similarly there are also symbols I have never seen in a closed caption file like: `"->> "` which I assume also denote a new speaker.

cc @isaacna @Shak2000 @dphoria any of you free to take this one?

Shak2000:
### Link to Relevant Issue

This pull request resolves #203 

### Description of Changes

Removing `>>` from the transcripts


dphoria (commented):
I'm OK with this, @evamaxfield . But in complete honesty, it's possible I did not understand the full intent for this issue; i.e. does this meet that fully.

dphoria:
@Shak2000 , given [this comment from To](https://github.com/CouncilDataProject/cdp-backend/issues/203#issuecomment-1242882487), I think this PR can change. i.e., It's possible the change should be to just update the regex pattern defined in `WebVTTSRModel.__init__()`.

evamaxfield:
Agree with @dphoria, can you try adding the optional regex and see if it works?

tohuynh:
I think to fix the mentioned issue, you'd need to do more than patch the code by removing the `>>` in the normalization step. 

Currently, the code is supposed to remove any `>>` (new speaker turn pattern) from the final transcript. See https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/sr_models/webvtt_sr_model.py#L87.

There is a clear bug that the code is not finding all the `>>` in the vtt file because some are present in the Oakland transcript.

You'd need to investigate what the [`captions`](https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/sr_models/webvtt_sr_model.py#L77) look like for an Oakland vtt file
```
# Read the caption file
captions = webvtt.read(file_uri)
```

And then investigate how the `captions` are processed in `_get_speaker_turns`.

Shak2000:
@tohuynh Good catch! After doing some more investigating, it seems that there is already a configuration for each CDP instance to define `caption_new_speaker_turn_pattern`. I tested it locally, and all we need seems to be to configure it as `>`. I will drop this pull request and create a pull request for Oregon.

PR #208, opened 2022-09-10
feature/add-download-and-upload-for-clips

evamaxfield:
Add the download from session id and the upload and gen url function in preparation for clipping.


cdp_backend/file_store/functions.py
```diff
@@ -172,3 +172,43 @@ def remove_local_file(filepath: Union[str, Path]) -> None:
     fs.rm(str(filepath))
 
     log.debug(f"Removed {filepath} from local file system.")
+
+
+def upload_file_and_return_link(
+    credentials_file: str,
```
dphoria, line 178: Completely fine; maybe `Path` type or `PathLike` may be better in the future?

dphoria (approved):

PR #214, opened 2022-09-23
feature/filter-out-invalid-captions

### LINKED ISSUE #213: Filter out bad caption files
### Feature Description

```Json
{"generator": "CDP WebVTT Conversion -- CDP v3.2.3", "confidence": 0.97, "session_datetime": "2022-09-14T10:00:00-04:00", "created_datetime": "2022-09-17T15:51:32.537666", "sentences": [{"index": 0, "confidence": 0.97, "start_time": 0.0, "end_time": 0.0, "words": [], "text": "", "speaker_index": 0, "speaker_name": null, "annotations": null}], "annotations": null}
```

Note `end_time` in the above output from a PIttsburgh, PA captions processing. Turns out that event does have a caption file in the Legistar data structure for the event _but_ it is in fact empty.

We want to filter these out just attempt speech-to-text.

### Use Case

Allow for proper transcript generation in `generate_transcript()` by appropriately filtering out empty caption files.

### Solution

Compare the lengths of the video and the caption file. If they differ by more than some threshold, e.g. 20%, throw away the caption file.
This means the scraper can no longer just hand off the caption file URL as-is.

### Alternatives

- Throw away caption files with file size less than some threshold, e.g. 100 bytes.
- Throw away caption files less than ~1 minute.

dphoria:
### Link to Relevant Issue

This pull request resolves #213 

### Description of Changes

In `create_event_gather_flow()`, in addition to calling `resource_exists()` for `session.caption_uri`, further validate the caption file by comparing its length against the video at `session.video_uri`. Reject, i.e. just do speech-to-text, if their lengths differ by more than 20%.

### DIFF
--- a/cdp_backend/pipeline/event_gather_pipeline.py
+++ b/cdp_backend/pipeline/event_gather_pipeline.py
@@ -160,7 +160,11 @@ def create_event_gather_flow(
                 if session.caption_uri is not None:
                     # If the caption doesn't exist, remove the property
                     # This will result in Speech-to-Text being used instead
-                    if not resource_exists(session.caption_uri):
+                    if not resource_exists(
+                        session.caption_uri
+                    ) or not file_utils.caption_is_valid(
+                        tmp_video_filepath, session.caption_uri
+                    ):
                         log.warning(
                             f"File not found using provided caption URI: "
                             f"'{session.caption_uri}'. "
--- a/cdp_backend/tests/utils/test_file_utils.py
+++ b/cdp_backend/tests/utils/test_file_utils.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import Optional
 from unittest import mock
 
+import ffmpeg
 import imageio
 import pytest
 from py._path.local import LocalPath
@@ -16,6 +17,7 @@ from cdp_backend.utils import file_utils
 from cdp_backend.utils.file_utils import (
     MAX_THUMBNAIL_HEIGHT,
     MAX_THUMBNAIL_WIDTH,
+    caption_is_valid,
     resource_copy,
 )
 
@@ -312,3 +314,26 @@ def test_clip_and_reformat_video(
     assert outfile.exists()
     assert outfile == expected_outfile
     os.remove(outfile)
+
+
+@pytest.mark.parametrize(
+    "video_uri, caption_uri, end_time, expected",
+    [
+        (EXAMPLE_VIDEO_FILENAME, "boston_captions.vtt", 120, False),
+        (EXAMPLE_VIDEO_FILENAME, "boston_captions.vtt", 60, True),
+    ],
+)
+def test_caption_is_valid(
+    resources_dir: Path, video_uri: str, caption_uri: str, end_time: int, expected: bool
+) -> None:
+    temp_video = "caption-test.mp4"
+    ffmpeg.input(str(bytes(resources_dir / video_uri), encoding="utf-8")).output(
+        temp_video, codec="copy", t=end_time
+    ).run(overwrite_output=True)
+
+    valid = caption_is_valid(
+        temp_video,
+        str(bytes(resources_dir / caption_uri), encoding="utf-8"),
+    )
+    os.remove(temp_video)
+    assert valid == expected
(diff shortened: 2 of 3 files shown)


cdp_backend/utils/file_utils.py
```diff
@@ -626,3 +628,40 @@ def clip_and_reformat_video(
...
+    Notes
+    -----
+    Duration of the video at video_uri
+    and the duration of the caption file are compared.
+    The caption file is accepted if the durations differ by no more than 20%.
+    """
+    try:
+        ffprobe = ffmpeg.probe(video_uri)
+    except ffmpeg.Error as e:
+        log.warning(f"ffprobe{video_uri}): {e.stderr}")
```
dphoria, line 658: typo. -> `f"ffprobe({video...`

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -160,7 +160,11 @@ def create_event_gather_flow(
                 if session.caption_uri is not None:
                     # If the caption doesn't exist, remove the property
                     # This will result in Speech-to-Text being used instead
-                    if not resource_exists(session.caption_uri):
+                    if not resource_exists(
+                        session.caption_uri
+                    ) or not file_utils.caption_is_valid(
```
dphoria, line 165: Not the correct call site. At this point this is almost always a remote resource.

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -312,3 +314,26 @@ def test_clip_and_reformat_video(
...
+    "video_uri, caption_uri, end_time, expected",
+    [
+        (EXAMPLE_VIDEO_FILENAME, "boston_captions.vtt", 120, False),
+        (EXAMPLE_VIDEO_FILENAME, "boston_captions.vtt", 60, True),
+    ],
+)
+def test_caption_is_valid(
+    resources_dir: Path, video_uri: str, caption_uri: str, end_time: int, expected: bool
+) -> None:
+    temp_video = "caption-test.mp4"
```
dphoria, line 329: Probably should use `TemporaryFile`.

cdp_backend/utils/file_utils.py
```diff
@@ -626,3 +634,50 @@ def clip_and_reformat_video(
...
+            caption_length = webvtt.read(local_caption_path).total_length
+        except (InvalidCaptionsError, MalformedCaptionError, MalformedFileError) as e:
+            log.warning(f"webvtt.read({caption_uri}): {str(e)}")
+            return False
+
+        similar_audio_streams = filter(
+            lambda s: s.get("codec_type", "") == "audio"
+            and math.isclose(
+                float(s.get("duration", "0.0")), caption_length, rel_tol=0.2
+            ),
```
evamaxfield, line 680: no change needed just a fun fact (if i am recalling it correctly) -- numpy.allclose (or some similar function) gets a different result than math.allclose and because it has been broken since like ~v0.4 or something they cant really change it because it would break so much code. LOL

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -312,3 +315,43 @@ def test_clip_and_reformat_video(
...
+def test_caption_is_valid(
+    resources_dir: Path,
+    video_uri: str,
+    caption_uri: str,
+    end_time: int,
+    is_resource: bool,
+    expected: bool,
+) -> None:
+    def path_as_str(path: Path) -> str:
+        return str(bytes(path), encoding="utf-8")
```
evamaxfield, line 346: Only question for this PR is "what is this doing???"

evamaxfield: you should just be able to call `str(Path("some-path.tiff"))` no need to cast to bytes

evamaxfield (approved):
Looks great! Thanks for getting to this!!!

One minor question / weird cast to bytes question but otherwise totally good to merge!

cdp_backend/tests/utils/test_file_utils.py
dphoria: Thank you for pointing this out. Until I thought the only way to get a string rep of the `Path` object was via `bytes()`. Turns out I misread the doc many months ago LOL.

PR #219, opened 2022-10-17
feature/firebase-to-sqlite: Create test script for Albuquerque

emilcecarlisa:
### Link to Relevant Issue

This pull request resolves #

### DIFF
--- a/cdp_backend/database/constants.py
+++ b/cdp_backend/database/constants.py
@@ -1,6 +1,23 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+BODY = "body"
+EVENT = "event"
+EVENT_MINUTES_ITEM = "eventMinutesItem"
+EVENT_MINUTES_ITEM_FILE = "eventMinutesItemFile"
+FILE = "file"
+MATTER = "matter"
+MATTER_FILE = "matterFile"
+MATTER_SPONSOR = "matterSponsor"
+MATTER_STATUS = "matterStatus"
+MINUTES_ITEM = "minutesItem"
+PERSON = "person"
+ROLE = "role"
+SESSION = "session"
+SEAT = "seat"
+TRANSCRIPT = "transcript"
+VOTE = "vote"
+
 
 class Order:
     ASCENDING = "ASCENDING"
--- a/cdp_backend/export-to-sqlite.py
+++ b/cdp_backend/export-to-sqlite.py
@@ -1,0 +1,246 @@
+from cdp_data import CDPInstances
+from cdp_data.utils import connect_to_infrastructure
+from fireo.models import Model
+from google.cloud.firestore_v1.transaction import Transaction
+from sqlalchemy import create_engine
+from sqlalchemy.orm import Session
+
+from cdp_backend.database import models as db_models
+from cdp_backend.database import sqlite_models
+from cdp_backend.database.constants import (
+    BODY,
+    EVENT,
+    EVENT_MINUTES_ITEM,
+    EVENT_MINUTES_ITEM_FILE,
+    FILE,
+    MATTER,
+    MATTER_FILE,
+    MATTER_SPONSOR,
+    MATTER_STATUS,
+    MINUTES_ITEM,
+    PERSON,
+    ROLE,
+    SEAT,
+    SESSION,
+    TRANSCRIPT,
+    VOTE,
+)
+
+# Init transaction and auth
+# fireo.connection(from_file=".keys/cdp-albuquerque-1d29496e.json")
+# Connect to the database
+connect_to_infrastructure(CDPInstances.Albuquerque)
+engine = create_engine("sqlite:///test.db", echo=True, future=True)
+
+model_list = [
+    {BODY: db_models.Body},
+    {EVENT: db_models.Event},
+    {EVENT_MINUTES_ITEM: db_models.EventMinutesItem},
+    {EVENT_MINUTES_ITEM_FILE: db_models.EventMinutesItemFile},
+    {FILE: db_models.File},
+    {MATTER: db_models.Matter},
+    {MATTER_FILE: db_models.MatterFile},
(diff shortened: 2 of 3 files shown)


cdp_backend/export-to-sqlite.py
```diff
@@ -0,0 +1,280 @@
...
+    for model in model_list:
+        name = model.get(MODEL_NAME)
+        print(f"STARTING INSERTS on modelname: {name}")
+
+        collection_iter = model.get(ACTUAL_MODEL).collection.fetch()
+
+        sql_schema = None
+        doc_idx = 0
+        for docRef in collection_iter:
+            # if element in doc is first
```
emilcecarlisa, line 262: Not doing anything with this yet but intend to fix this so that we're not calling `get_sql_schema` on every row in collection

cdp_backend/export-to-sqlite.py
```diff
@@ -0,0 +1,276 @@
...
+        return sqlite_models.Vote()
+    else:
+        return
+
+
+def get_schema_properties(sql_schema: Model, doc: Transaction) -> Model:
+    model = sql_schema
+    model.id = doc.id
+
+    if model.get_class_by_tablename() == type(sqlite_models.Body()):
```
evamaxfield, line 151: A lot of this can be dynamic.

Calling `dir()` on any Python object will let you see what attributes it has available so may want to try that for fun but also I should have told you that there is a way to get the firestore model field list:

```python
from cdp_backend.database.models import Body  # just an example

fields = Body._meta.field_list
```

`fields` now is the following dictionary:

```
{'name': <fireo.fields.text_field.TextField at 0x7f80bcc1b940>,
 'description': <fireo.fields.text_field.TextField at 0x7f80bcc1b6a0>,
 'start_datetime': <fireo.fields.datetime_field.DateTime at 0x7f80bcc1b7c0>,
 'end_datetime': <fireo.fields.datetime_field.DateTime at 0x7f80bcc3c040>,
 'is_active': <fireo.fields.boolean_field.BooleanField at 0x7f80bcc3c100>,
 'external_source_id': <fireo.fields.text_field.TextField at 0x7f80bcc3c1c0>}
```

I assume that you should be able to do some sort of matching between the firestore model and the sqlite model since they have the same attr names:

```python
from cdp_backend.database import DATABASE_MODELS
from cdp_backend.database import sqlite_models

for firestore_model in DATABASE_MODELS:
    # this is a lookup in the sqlite_models module
    # for whatever object has the same class name as the current
    # iterating firestore model
    # i.e. File, or Body, or EventMinutesItem, etc.
    matching_sqlite_model = getattr(sqlite_models, firestore_model.__name__)
    for field in firestore_model._meta.field_list:
        setattr(
            matching_sqlite_model,
            field,
            getattr(firestore_model, field)
        )     
```

I am just trying to find places to reduce the amount of copy pasted code.

evamaxfield (commented):
A lot of my comments are basically just "python secrets" of how to access object attributes.

Only major comment is I would love to see the `export-to-sqlite.py` main functionality be moved to a function in the library. `export_database_to_sqlite` that takes the database name as a parameter?

evamaxfield:
Really really good work so far! Sorry that I have been a bit AFK

cdp_backend/export-to-sqlite.py
```diff
@@ -0,0 +1,276 @@
...
+        model.name = doc.name
+        model.description = doc.description
+        model.start_datetime = doc.start_datetime
+        model.is_active = doc.is_active
+        model.external_source_id = doc.external_source_id
+    elif model.get_class_by_tablename() == type(sqlite_models.Event()):
+        model.body_id = doc.body_ref.get().id
```
evamaxfield, line 158: Responding to your specific question here (great question by the way).

There is a method for doing this _without_ fetching the document. The reference has the document id and you can get with more property access:

```python
# Just example setup code
from cdp_data import CDPInstances
from cdp_data.utils import connect_to_infrastructure
from cdp_backend.database.models import Event
connect_to_infrastructure(CDPInstances.Seattle)

# Actual model + getting the ref id
e = list(Event.collection.fetch(1))[0]
print(e.body_ref.ref.id)
```

cdp_backend/export-to-sqlite.py
```diff
@@ -0,0 +1,276 @@
...
+)
+
+# Init transaction and auth
+# fireo.connection(from_file=".keys/cdp-albuquerque-1d29496e.json")
+# Connect to the database
+connect_to_infrastructure(CDPInstances.Albuquerque)
+engine = create_engine("sqlite:///test.db", echo=True, future=True)
+
+# Create tables!
+sqlite_models.Base.metadata.create_all(engine, tables=None, checkfirst=True)
```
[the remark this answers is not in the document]
emilcecarlisa: Yes, it enables the “check first for the table existing” logic

cdp_backend/database/constants.py
```diff
@@ -1,6 +1,23 @@
...
+MATTER_FILE = "matterFile"
+MATTER_SPONSOR = "matterSponsor"
+MATTER_STATUS = "matterStatus"
+MINUTES_ITEM = "minutesItem"
+PERSON = "person"
+ROLE = "role"
+SESSION = "session"
+SEAT = "seat"
+TRANSCRIPT = "transcript"
+VOTE = "vote"
```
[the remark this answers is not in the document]
emilcecarlisa: I'm using Python's [inspect module](https://docs.python.org/3/library/inspect.html) to get the name of the db classes and I'm getting back ' _SpecialGenericAlias' or 'ModelMeta' for the model.__class__ or model.__class__.__name__ respectively. Seems like since we inherit from both Model and Meta is a subclass (?) that's where 'ModelMeta' comes from. Not sure what ' _SpecialGenericAlias' is, but I'll continue to look at that. As for calling 'model.__name__', we have to assign a variable 'name' to get anything returned from that

PR #221, opened 2022-11-08
Add transcription range fields to database and ingestion models, add …

### LINKED ISSUE #211: Accept timestamps and only process a subset of a video as an event
### Feature Description

The ability to transcribe only a portion of a video and have it be considered an event. 

### Use Case

We are scraping a state legislature site for bill hearings and many of the videos are ~5-6 hours long with discussions about several bills in one video. The timestamps of when the bill discussions happen are programmatically available. We want to be able to use the same video multiple times, but with different timestamp ranges, so the result is several Events each only with a slice of the same mp4, and no duplicate processing happens.

### Solution

Add something like a start_timestamp / stop_timestamp to the `EventIngestionModel`. When these values are present for the event, the backend should only process and transcribe the given video & audio range.

The event frontend should also only render this segment of the whole video (potentially covered by recent hackathon efforts?)

chrisjkhan:
### Link to Relevant Issue

This pull request resolves #211 

### Description of Changes

-Add optional transcription range fields to database model
-Add optional transcription range fields to ingestion model
-Add validator for time duration
-Add range filter to ffmpeg audio split (similar to clip functionality)
-Update tests for valid inputs


evamaxfield (commented):
Codewise, this all looks great. Would be curious as to how you found or contributing notes and all of that. Any places for improvement?

Implementation wise... did you have a chance to test this on a dev deployment? I think this would _transcribe_ the portion selected but the web UI / session interface may fail because the video doesn't match up with the audio... the best way to do this interestingly may be to split the video by timestamp and then do all the normal downstream processing rather than split the audio by timestamp...

Are you on the call this Thursday? If so it might be good to talk about this feature then. / How to dev-deploy

chrisjkhan:
> Codewise, this all looks great. Would be curious as to how you found or contributing notes and all of that. Any places for improvement?
>
I made it this far! If there's anything that I'm doing that makes it appear I missed the documentation, let me know and I might have some feedback. 
>
> Implementation wise... did you have a chance to test this on a dev deployment?
>
No, that was the last thing I mentioned to @smai-f. Hopefully, we can get a test process going as we integrate the scraper.
>
>I think this would _transcribe_ the portion selected but the web UI / session interface may fail because the video doesn't match up with the audio... the best way to do this interestingly may be to split the video by timestamp and then do all the normal downstream processing rather than split the audio by timestamp...
>
My ideal was to do as you mentioned, and probably integrate with clip functionality. The main reason I didn't was because there seems to be a reliance on existing video hosting, and I thought that would be worth preserving.

I didn't realize how the front end functionality worked (oops!), but I think there are some easy ways to preserve both the existing video hosting and accurate timestamps. The simplest might be to send the offset to the transcribe function to alter the timestamps in the transcript: https://github.com/CouncilDataProject/cdp-backend/blob/1e5b80743c4c0374f5a7d7df9731346d61f02f55/cdp_backend/sr_models/google_cloud_sr_model.py#L199

In addition, we can update the front end so it seeks to the appropriate place on page load: https://github.com/CouncilDataProject/cdp-frontend/blob/77944572b816ed07ef0dcffd772fc19b52aea12e/src/components/Details/EventVideo/EventVideo.tsx#L133
>
> Are you on the call this Thursday? If so it might be good to talk about this feature then. / How to dev-deploy
>
I should be available. Can you add me to the event?

cdp_backend/database/validators.py
```diff
@@ -72,6 +72,34 @@ def router_string_is_valid(router_string: Optional[str]) -> bool:
...
+    Returns
+    -------
+    status: bool
+        The validation status.
+    """
+    if time_duration is None:
+        return True
+
+    # HH:MM:SS
+    if re.match(r"^((((\d{1,2}:)?[0-5])?\d:)?[0-5])?\d$", time_duration):
```
dphoria, line 97: Does this mean `time_duration="00:00:00"` would yield `True` from this method?

cdp_backend/utils/file_utils.py
```diff
@@ -282,7 +288,11 @@ def split_audio(
         raise IsADirectoryError(resolved_audio_save_path)
 
     # Construct ffmpeg dag
-    stream = ffmpeg.input(resolved_video_read_path)
+    stream = ffmpeg.input(
+        resolved_video_read_path,
+        ss=start_time or "0",
```
chrisjkhan, line 293: Default `start_time` equal to 00:00:00

cdp_backend/database/validators.py
chrisjkhan: That's the intention. The default `start_time` is equal to `00:00:00`, so it's valid for the type in ffmpeg. Maybe we want additional manipulation or verification for `end_time` or duration pairs, but leaving it this way seems reasonable to me. If for some reason someone wants to exclude transcription for a particular video, this would be a method of accomplishing that. I don't know how the full stack is going to behave with these edge cases, but I didn't see much edge case testing with media files, so maybe it's OK to just do some manual tests?

cdp_backend/tests/database/test_validators.py
```diff
@@ -48,6 +48,39 @@ def test_router_string_is_valid(router_string: str, expected_result: bool) -> No
...
+        ("1:11", True),
+        ("11:11", True),
+        ("1:11:11", True),
+        ("99:59:59", True),
+        ("0", True),
+        ("00", True),
+        ("0:00", True),
+        ("00:00", True),
+        ("0:00:00", True),
+        ("00:00:00", True),
```
chrisjkhan, line 66: Add validation test cases for `0` durations

cdp_backend/utils/file_utils.py
dphoria: Looks like we need to check the duration. i.e. I tried the following (not using `ffmpeg-py` or whatever python package we are using; `ffmpeg` directly but I presume the behavior may be similar)

```
$ ffmpeg -i cdp-backend/cdp_backend/tests/resources/example_video.mp4 -ss 00:00 -to 00:00 -codec copy out.mp4
-to value smaller than -ss; aborting.
```

Obviously, the actual video file is irrelevant. The point is that it _looks_ like duration must be positive.

dphoria: I did try the same command with `-to 00:01` and that did succeed.

dphoria:
Beyond what Eva already mentioned, about the different video and audio stream lengths maybe being an issue, I think my only wish is that `split_audio` checks `end_time > start_time` in some way.

Thank you very much @chrisjkhan for this!

cdp_backend/utils/file_utils.py
chrisjkhan: FFmpeg was throwing errors during tests, so I assumed test cases would be enough of an indicator. It turns out I had a small typo that prevented the test cases from working properly, along with only file utils tests using an ffmpeg mock.

After the Thursday meeting we decided to add event gather validation to throw errors at that time.

chrisjkhan:
> Beyond what Eva already mentioned, about the different video and audio stream lengths maybe being an issue, I think my only wish is that `split_audio` checks `end_time > start_time` in some way.
> 
We decided at the meeting to do full video clipping and host the video. A combination of reusable functionality, relative low priority of original video hosting, simplicity of downstream issues, and overall UX. I'll change the functionality and the semantics of the whole issue.
>
> Thank you very much @chrisjkhan for this!
>
You're welcome

dphoria:
> > Beyond what Eva already mentioned, about the different video and audio stream lengths maybe being an issue, I think my only wish is that `split_audio` checks `end_time > start_time` in some way.
> 
> We decided at the meeting to do full video clipping and host the video. A combination of reusable functionality, relative low priority of original video hosting, simplicity of downstream issues, and overall UX. I'll change the functionality and the semantics of the whole issue.
> 
Sounds great. Just let me know / hit the "re-request review" button whenever.

chrisjkhan:
I may have gone a little overboard with changes trying to slightly increase consistency across method signatures. The only thing I changed that might be controversial is making the hash calculation part of an already large task rather than keeping it separate. If you really want the hash to be performant, I would suggest modifying it to use CRC/MD5/SHA-1 and do a combination of the first x bytes/blocks and file size.

I tested negative duration pairs (`start_time < end_time`) causing a event processing failure, as desired. Positive duration pairs cause the file to be trimmed and CDP hosted, and considered unique/distinct from full length videos. Transcription works as expected.

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -80,6 +80,7 @@ def test_hash_file_contents(tmpdir) -> None:  # type: ignore
     assert hash_a != hash_b
 
 
+@pytest.mark.parametrize("video_filename", [(EXAMPLE_VIDEO_FILENAME)])
```
dphoria, line 83: Just curious, what is the reason this is not included in the existing parametrization below?

dphoria: Oh I see, this gets used as a parameter for other tests too?

dphoria: Oh, it's because the count is different; this has 1 but the next `parametrize()` has 5?

cdp_backend/utils/file_utils.py
```diff
@@ -681,4 +708,13 @@ def caption_is_valid(video_uri: str, caption_uri: str) -> bool:
...
             ffprobe.get("streams", []),
         )
+        audio_streams = filter(
+            lambda s: s.get("codec_type", "") == "audio",
+            ffprobe.get("streams", []),
+        )
+        for s in audio_streams:
+            duration = s.get("duration", "0.0")
+            log.warning(f"stream length {duration}")
+        log.warning(f"caption length {caption_length}")
```
dphoria, line 718: Are these true-positive warnings? i.e. Should they stay as `warning()`, or be changed to e.g. `info()` or `debug()`?

cdp_backend/pipeline/ingestion_models.py
```diff
@@ -139,9 +139,22 @@ class Session(IngestionModel, DataClassJsonMixin):
...
     external_source_id: Optional[str] = None
 
+    def __post_init__(self) -> None:
+        if self.video_start_time and self.video_end_time:
+            start = list(map(int, ("00:00:0" + self.video_start_time).split(":")))
+            end = list(map(int, ("00:00:0" + self.video_end_time).split(":")))
+            start.reverse()
+            end.reverse()
+            start_seconds = start[0] + start[1] * 60 + start[2] * 3600
+            end_seconds = end[0] + end[1] * 60 + end[2] * 3600
```
dphoria, line 154: Haha I like this. I'm not sure I would have thought of this. Pretty neat.

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -116,6 +119,10 @@ def test_split_audio(  # type: ignore
             assert str(tmp_dir_audio_save_path.with_suffix(".out")) == stdout_log
             assert str(tmp_dir_audio_save_path.with_suffix(".err")) == stderr_log
 
+            os.remove(audio_file)
+            os.remove(stdout_log)
+            os.remove(stderr_log)
```
dphoria, line 124: So, if there in exception before this, these files would get left behind, right? This is a test, so I really don't care that hard; just something that I noticed.

cdp_backend/tests/utils/test_file_utils.py
chrisjkhan: This was an example of trying to get test patterns more consistent. I don't know if it helped. As to why they're not combined into one `parameterize` annotation, yeah, because this parameter has one value and the next has five. It's more descriptive and concise to keep it as a separate annotation. This strategy is used elsewhere quite a bit.

cdp_backend/utils/file_utils.py
chrisjkhan: I had a feeling I forgot to correct some log statements. This should be changed, at least, if not removed.

cdp_backend/pipeline/ingestion_models.py
chrisjkhan: Stretching the bounds of readability. I should have added a comment.

cdp_backend/tests/utils/test_file_utils.py
chrisjkhan: I cared just hard enough to clean up where there was no cleanup before. I roughly followed the pattern in the rest of the file. I didn't care hard enough to improve them all. I like the idea though.

dphoria (approved):
I think I'm good with this. Thank you, Chris! I'll let Eva do the honors, though. She's a much better reviewer.

smai-f:
Thanks everybody! Here are some manual tests using Chris' latest commit in the Montana legislature instance:

- [random DT](https://www.openmontana.org/montana-legislature-council-data-project/#/events/98fe24e6a8bc)
- [different portion of same meeting](https://www.openmontana.org/montana-legislature-council-data-project/#/events/ea2d8dcf4fdd)
- [different portion earlier than both the above](https://www.openmontana.org/montana-legislature-council-data-project/#/events/b2ae9c217781)
- [0 start](https://www.openmontana.org/montana-legislature-council-data-project/#/events/d148d814062d)
- [no end](https://www.openmontana.org/montana-legislature-council-data-project/#/events/eb7a988f9a51)
- [start after end](https://github.com/OpenMontana/montana-legislature-council-data-project/actions/runs/3556660076/jobs/5974181880)

evamaxfield (commented):
This all looks good to me. I have one remaining question before I click the approve and merge buttons.

Q: should we take the trim start time and add that to the session start time / event start times?

lets say the event starts at noon but you want to trim to 1 hour into the video. should the session and event start time say: "1pm"?

evamaxfield:
Also, thank you for such great work on this! Really nicely done and thank you @smai-f for helping test!

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -330,19 +324,41 @@ def convert_video_and_handle_host(
...
-        # Remove old mkv file
         fs_functions.remove_local_file(video_filepath)
 
         # Update variable name for easier downstream typing
-        video_filepath = mp4_filepath
+        video_filepath = str(mp4_filepath)
+
+    # host trimmed videos because it's simpler than setting
+    # up transcription and playback ranges
+    elif trim_video:
```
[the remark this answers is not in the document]
chrisjkhan: There may be a more clear way to set up the control logic, but there is some interaction with the conversion logic. If the video needs to be converted, the trimming can be done there, so trimming can be skipped elsewhere. Trim and convert should probably just be combined into one block, especially since the utility function was combined: `if trim or convert: trim_and_convert() elif: ...`. I was hesitant to change it too much. Your call.

cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -406,6 +425,10 @@ def split_audio(
         The unique identifier for the session.
     tmp_video_filepath: str
         The local path for video file to generate a hash for.
+    start_time: str
+        The start time of the clip in HH:MM:SS.
+    end_time: str
+        The end time of the clip in HH:MM:SS.
```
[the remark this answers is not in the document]
chrisjkhan: Updating...

evamaxfield:
> lets say the event starts at noon but you want to trim to 1 hour into the video. should the session and event start time say: "1pm"?

cc @BrianL3 @dphoria @smai-f this is a question i would love all of your opinions on

cdp_backend/pipeline/event_gather_pipeline.py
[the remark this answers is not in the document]
evamaxfield: Ah I see. Okay, this works for me. Only one question remaining which is in the general PR. nothing on current code changes

chrisjkhan:
> lets say the event starts at noon but you want to trim to 1 hour into the video. should the session and event start time say: "1pm"?

Ah, I missed this comment. My guess is it should still be noon. There is a `session_datetime`, so that value would be over constrained. I chose the very general semantics of `video_start_time` to leave room for any possible reasons for trimming, leaving a disconnect between event/session/etc times and video times. People more familiar with usage would know better though. If we did make this kind of link in timing, it might make sense to narrow the semantics.

The only case I can imagine is a session "starting" at noon, the video recording starting at noon, but then actual activity starting at 1pm. My guess is, if this 1 hour of dead time is in the source data being scraped, `session_datetime` would reflect the actual start time instead of the false start time. If dead time not in the source data, it couldn't be scraped and then trimmed.

dphoria:
> > lets say the event starts at noon but you want to trim to 1 hour into the video. should the session and event start time say: "1pm"?
> 
> Ah, I missed this comment. My guess is it should still be noon. There is a `session_datetime`, so that value would be over constrained. I chose the very general semantics of `video_start_time` to leave room for any possible reasons for trimming, leaving a disconnect between event/session/etc times and video times. People more familiar with usage would know better though. If we did make this kind of link in timing, it might make sense to narrow the semantics.
> 
> The only case I can imagine is a session "starting" at noon, the video recording starting at noon, but then actual activity starting at 1pm. My guess is, if this 1 hour of dead time is in the source data being scraped, `session_datetime` would reflect the actual start time instead of the false start time. If dead time not in the source data, it couldn't be scraped and then trimmed.

First, great question Eva. :clap: I'm in line with Chris' thoughts; (I hope I'm understanding correctly) At the core, there are 2 fields aded to the model, the video start and end times (right?). At least to me, if I look at a model, and see that 'session start time' is noon, and 'video start time' is 1 PM, that would sound pretty natural. I may at first be slightly confused, but even without any external explanation, I think I would realize that this means that in this case, the video begins 1 hour into the session.

If I'm looking at an example of what's probably the vast majority of the models - 'session start time' == 'video start time', that again wouldn't throw me off, especially if I have at that point, seen an example like above when they are different.

chrisjkhan:
>At least to me, if I look at a model, and see that 'session start time' is noon, and 'video start time' is 1 PM, that would sound pretty natural.

Maybe there is still some confusion. `video_start_time` is an `HH:MM:SS` time/duration, rather than a datetime. It's relative to the beginning of the video. Should I do something to make this more clear?

dphoria:
> > At least to me, if I look at a model, and see that 'session start time' is noon, and 'video start time' is 1 PM, that would sound pretty natural.
> 
> Maybe there is still some confusion. `video_start_time` is an `HH:MM:SS` time/duration, rather than a datetime. It's relative to the beginning of the video. Should I do something to make this more clear?

Thank you for pointing that out. I missed it!

chrisjkhan:
> I would just put it in the class docstring...

Added. Let me know how it looks. I followed the `Notes` model in the rest of the file. It's a bit verbose.

evamaxfield (approved):
Perfect. Verbose is more than welcome. Clicking the merge button!

PR #225, opened 2022-12-28
bugfix/unique-temp-filenames

### LINKED ISSUE #226: Cannot store and transcribe multiple videos that are clipped via the video_start_time and video_end_time event params
### Describe the Bug

I’m scraping multiple events that have `video_start_time` and `video_end_time` params to only transcribe a subset of a full video. 

The call to `clip_and_reformat_video` when clipping the subset of a video isn’t passing in an `outpath_path` param, so they are all named the default `clipped.mp4` which is [erroring out](https://github.com/OpenMontana/montana-legislature-council-data-project/actions/runs/3783141951/jobs/6431445321) from the duplication.

### Expected Behavior

Scraping multiple videos at once with trimmed segments should work.

### Reproduction

Event list printed near top of logs: https://github.com/OpenMontana/montana-legislature-council-data-project/actions/runs/3783141951/jobs/6431445321

### Environment

-   OS Version: _[e.g. macOS 11.3.1]_ Github Actions
-   cdp-backend Version: _[e.g. 0.5.0]_ 3.2.10

chrisjkhan:
This pull request resolves #226 

Use content hash for temp video file naming to prevent collisions across


cdp_backend/pipeline/event_gather_pipeline.py
```diff
@@ -127,8 +128,11 @@ def create_event_gather_flow(
         for event in events:
             session_processing_results: List[SessionProcessingResult] = []
             for session in event.sessions:
-                # Download video to local copy
-                resource_copy_filepath = resource_copy_task(uri=session.video_uri)
+                # Download video to local copy making
+                # copy unique in case of shared session video
+                resource_copy_filepath = resource_copy_task(
+                    uri=session.video_uri, dst=f"{str(uuid4())}_temp.mp4"
```
dphoria, line 134: Are we guaranteed to have an MP4 video at this point? i.e. Is it safe to specify the `.mp4` suffix?

dphoria (commented):
Thank you Chris! Just one question.

cdp_backend/pipeline/event_gather_pipeline.py
chrisjkhan: I need to make an additional change where the suffix is taken from the downloaded file. For `youtube_copy` etc, the suffix is forced.

cdp_backend/tests/utils/test_file_utils.py
```diff
@@ -35,6 +35,58 @@
...
+    os.remove(new_path)
+
+
+@pytest.mark.parametrize(
+    "path, addition, expected_result",
+    [
+        (Path("file.ext"), "-new", "file-new.ext"),
+    ],
+)
+def test_rename_append_to_stem(path: Path, addition: str, expected_result: str) -> None:
```
evamaxfield, line 81: I really like this util func.

evamaxfield (commented):
I think this all looks good to me with two caveats:

1. the thing @dphoria already mentioned about pulling in the correct extension
2. a bit more complicated, one of the things that is decently important is that the produced files related to a session are all uploaded with deterministic names. I think a prior version of this PR actually used the `session_content_hash` which I think is the correct thing to use instead of a `uuid`. The reasoning here is that when it comes time to pulling down audio files from the database (which is never done for the app side of things but is done in my research) we need a deterministic name. If all of that made sense, the question is: after these changes, do the files that are ultimately stored to the file storage system / Google Cloud Storage have deterministic naming?

cdp_backend/tests/utils/test_file_utils.py
chrisjkhan: 1. I need to create a suffix override flag to preserve existing functionality, and then detect the existing suffix to transfer to the saved file. The reason this may be necessary is if parallel tasks are saving the same file with the same name to be modified in different ways, there will be a collision.
2. https://github.com/CouncilDataProject/cdp-backend/blob/74362559829fa1b49862b649eb50076932c57d10/cdp_backend/pipeline/event_gather_pipeline.py#L405
The CDP hosted file should still be deterministic. I just save the temp file with a UUID name so that it doesn't conflict with other sessions, especially if they have the same video file. A UUID is already used in this process, so I assume it's generally not a problem to have an unpredictable temp file name.
https://github.com/CouncilDataProject/cdp-backend/blob/74362559829fa1b49862b649eb50076932c57d10/cdp_backend/utils/file_utils.py#L208

chrisjkhan: https://github.com/CouncilDataProject/cdp-backend/blob/6b31f4b2921de98cd41e2cf53d3d347d68daec85/cdp_backend/utils/file_utils.py#L185

cdp_backend/pipeline/event_gather_pipeline.py
chrisjkhan: https://github.com/CouncilDataProject/cdp-backend/blob/6b31f4b2921de98cd41e2cf53d3d347d68daec85/cdp_backend/utils/file_utils.py#L185

evamaxfield (approved):
Assuming the correct extension gets passed through. Then all looks good. Thanks @chrisjkhan for checking the determinism aspect ❤️

Will merge once I see the extension work go through and tests pass.

evamaxfield (approved):
Looks good to me! @dphoria can you also check this. Will merge soon!

dphoria (approved):
Ship it~ :rocket:

chrisjkhan:
> Looks good to me! @dphoria can you also check this. Will merge soon!

@evamaxfield actually, it's not working right now. I was wanting to reach out to see if you knew what might be wrong. The temporary video file exists in the context of the `convert_video_and_handle_host` function, after the `resource_copy_task`, but it does not exist in the context of the `file_utils.clip_and_reformat_video` function, with no other commands called in between. I'm starting to think it's something small, like a typo. I thought maybe it had to do with the Prefect architecture, but I don't think so now. It is essentially the same architectural form as before any changes were made for trimming, so it seems like it must be a minor issue.

File exists here: https://github.com/CouncilDataProject/cdp-backend/blob/01595c3d6653649c4fab376ac686cfd5ada91f9a/cdp_backend/pipeline/event_gather_pipeline.py#L358
But not here: https://github.com/CouncilDataProject/cdp-backend/blob/01595c3d6653649c4fab376ac686cfd5ada91f9a/cdp_backend/utils/file_utils.py#L724

@dphoria let me know if you have any ideas

dphoria:
:rofl: I have failed as a reviewer. Anyway, good catch Chris. I'll see...

chrisjkhan:
Relevant logs: https://github.com/OpenMontana/montana-legislature-council-data-project/actions/runs/3811636069/jobs/6484420314#step:9:81

evamaxfield:
I will also try to take a look asap

cdp_backend/utils/file_utils.py
```diff
@@ -634,6 +713,20 @@ def clip_and_reformat_video(
     """
     import ffmpeg
 
+    output_path = output_path or rename_append_to_stem(video_filepath, "_clipped")
```
chrisjkhan, line 716: @evamaxfield  @dphoria typo indeed. `rename_` shouldn't be in there :disappointed:

chrisjkhan:
Looking good: https://github.com/OpenMontana/montana-legislature-council-data-project/actions/runs/3814172947/jobs/6488287009

cdp_backend/utils/file_utils.py
dphoria: Just curious, was/is there some way we could've caught this through a unit test?

evamaxfield:
Thanks for making all the changes @chrisjkhan!

Merging and releasing new version. Upgrade should rollout in ~2 hours.

cdp_backend/utils/file_utils.py
chrisjkhan: From a unit test, we could have checked a no `outfile` set of parameters. I just pushed a change for that test. I verified that it fails with the previous code and passes with the latest update.

evamaxfield: Can you make a new PR for just that test please! Sounds very useful!

chrisjkhan: https://github.com/CouncilDataProject/cdp-backend/pull/227

PR #229, opened 2023-03-05
docs/improve-dev-infra-setup

isaacna:
### Description of Changes

Some of the steps in the dev-infrastructure README were out of order, so rearranging them based on my recent setup experience. Also added:
- fixing an issue where `dev-cookiecutter-metadata.yaml` was not included as one of the copied files as part of `get_cdp_infrastructure_stack`, which would cause the setup to fail if there was a directory chosen besides the default `cdp-backend/dev-infrastructure`. 
- adding an optional `cookiecutter_yaml` param to the `setup-and-deploy` just command in case it's desired


evamaxfield (approved):
Minor comments / changes. Thanks!!!

Issue #81, opened 2021-08-14
Add MatterIndexPipeline


JacksonMaxfield:
### Feature Description

_A clear and concise description of the feature you're requesting._

Add matter index pipeline to stored indexed terms for matters.

### Use Case

_Please provide a use case to help us understand your request in context._

Search for matters based off of a plain text search query.

### Solution

_Please describe your ideal solution._

Similar to event index pipeline but instead of parsing the transcript parse the matter document text and link to `matter_ref` instead of `event_ref`.

### Alternatives

_Please describe any alternatives you've considered, even if you've dismissed them._

JacksonMaxfield:
Hey @Shak2000 @isaacna is available to help on this one if you have questions / need help with getting the pipeline set up :slightly_smiling_face:

sagarrat7:
I’m going to start working on this issue.

PR #231, opened 2023-04-04
feature/matter-text-extraction

sagarrat7:
### Link to Relevant Issue
Related to #81 

### Description of Changes
Adds utility function `parse_document()` that extracts text from docx, doc, pdf, and pptx matters to be used in indexing matters in addition to transcripts. 
Note: pptx files contain extra "Title" and "/docProps/thumbnail.jpeg". This can be removed if needed.


cdp_backend/utils/file_utils.py
```diff
@@ -739,3 +744,160 @@ def clip_and_reformat_video(
...
+    ----------
+    document_uri: str
+        The matter file uri.
+
+    Returns
+    -------
+    str:
+        A string of all text in the matter file.
+    """
+    response = requests.get(document_uri, stream=True)
```
evamaxfield, line 763: Nit: I think this open buffer can be a bit more safely handled using the context manager:

```python
with requests.get(document_uri, stream=True) as response:
    ...
```

It auto closes the open stream when done. At least that is how requests used to work.

cdp_backend/utils/file_utils.py
```diff
@@ -739,3 +744,160 @@ def clip_and_reformat_video(
...
+    if response.status_code != 200:
+        log.error(
+            "Request to " + document_uri + " failed with: " + str(response.status_code)
+        )
```
evamaxfield, line 767: I would use the `response.raise_for_status()` function.

evamaxfield (commented):
This is great!!! Thanks so much. A few minor requests nitpicks but otherwise good to go!

cdp_backend/utils/file_utils.py
evamaxfield: mmmmm yea. Mocking this behavior is a bit harder. I would say just ignore this one for now. It is minor and these files should never be too large!

evamaxfield (approved):
Looks good to me! Thank you!

Issue #230, opened 2023-04-03
Reduce complexity of event gather pipeline


evamaxfield:
cc @whargrove -- I know that the Montana legislature event gather pipeline has been failing for a bit, most commonly iirc from disk fill up. I have been trying to think of ways to reduce those failures from happening.

The original reason we used Prefect for our pipeline management was because we used to be able to process events in parallel (when we were using google cloud speech-to-text we could parallel process because they were doing the heavy lifting, now that we are using whisper and we only have a single GPU, it is easiest to do everything synchronously). The problem is, prefect naturally assumes things can be run in parallel (and partially that we have infinite resources) so it runs tasks in the a semi-breadth first manner. This is what is causing a lot of your out of memory issues -- prefect will download all of the videos in your batch, then move on to stripping the audio, then move on to transcribing, etc, prior to uploading / deleting the file from disk.

This sort of makes prefect useless.

I am opening this issue to hopefully have a discussion on reducing the complexity of the entire library / the main event processing pipeline by reducing our usage of prefect, or entirely removing it from the library.

---

#### Remake the pipeline to work on a single session

Right now, the pipeline takes in [any number of events](https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/event_gather_pipeline.py#L148) and creates session processing tasks for each session within each event. Each of these can run in parallel. If I were describe this potential change in pseudo code it would be from currently having this:

```python
with Flow("event-gather-pipeline") as flow:
    events = get_events_func()
    for event in events:
       for session in sessions:
           task1
           task2
           task3
```

to the following:

```python
events = get_events_func()
for event in events:
    for session in sessions:
        with Flow("processing-single-session") as flow:
            task1
            task2
            task3
            
# some kind of nice utility to aggregate errors
log(agg_errors)  # "3 / 8 sessions failed to process" or "all sessions processed successfully", etc.
```

This would naturally make it so one session is processed at a time (and all files should hopefully be deleted after each session is done) and we could actually re-enable some minor parallelism (things like thumbnail generation during audio processing).

#### Remove prefect entirely

This would remove the complexity A LOT, but it would also remove all the niceties like the free logging, the easy error management / recovery, and most especially, the retry and backoff system.

But it would entirely remove prefect and make our whole pipeline synchronous which is much less complex for sure.

### My opinion

[...]

PR #232, opened 2023-05-06
feature/reduce-event-gather-complexity

evamaxfield:
Had parts of this code sitting around. Finished it up in ~30 minutes. Need to test on a dev instance but pretty confident it will work.

Main change is resolving #230 -- moving from processing all sessions in a single flow to producing multiple flows, one for each session. I haven't touched any multiprocessing just yet to be safe.


evamaxfield:
Heading for the night, might get a chance to test this tomorrow. If it works on a dev instance will fast merge and deploy

PR #236, opened 2023-05-13
Use requests stream and shutil.copyfileobj to constrain memory usage during resource copy

### LINKED ISSUE #235: Inefficient usage of requests.get for very large videos causes event gather to fail
### Describe the Bug

Usage of [requests](https://github.com/CouncilDataProject/cdp-backend/blob/0982830358df2c62ab00a05bf8f862c4f3ee168c/cdp_backend/utils/file_utils.py#L237) to write the content to the open destination file as a BufferedWriter causes unconstrained memory usage based on the size of the video file being copied from the remote origin.

For example, consider the mp4 at https://sg001-media.sliq.net/00309/Media2/2023/04/27/House%20Floor%20Session-House%20Chamber-080505_46177_1113.mp4...

```sh
curl -Ss -I https://sg001-media.sliq.net/00309/Media2/2023/04/27/House%20Floor%20Session-House%20Chamber-080505_46177_1113.mp4 | grep ^content-length
content-length: 8068641879
```

The video is ~8GB in size and uses too much memory for n1-standard-4 machine type we use by default in Google Cloud especially when whisper model is already loaded.

### Expected Behavior

Downloading video files from remote origin should use a pattern that constrains memory usage in the python process.

### Reproduction

```python
import requests

with open("movie.mp4", "wb") as f:
    f.write(
        requests.get("https://sg001-media.sliq.net/00309/Media2/2023/04/27/House%20Floor%20Session-House%20Chamber-080505_46177_1113.mp4").content
    )
```

Running this will download the 8G file and copy to output file. The memory usage will be unconstrained since requests to will load the entire body into memory before copying the file to disk.

### Environment

- OS Version: N/A
- cdp-backend Version: latest

whargrove:
### Link to Relevant Issue

This pull request resolves #235

### Description of Changes

- When copying resources from a remote origin over HTTP(S) prefer to stream the response body and copy chunks into the destination file insead of loading the entire file into memory first before writing.
- Before the change, running resource_copy("uri") in python REPL would use unbounded rss (up to file size). After the change, running the same in python REPL uses ~300M rss.

### DIFF
--- a/cdp_backend/tests/utils/test_file_utils.py
+++ b/cdp_backend/tests/utils/test_file_utils.py
@@ -11,6 +11,7 @@ from unittest import mock
 
 import imageio
 import pytest
+import requests_mock
 
 from cdp_backend.utils import file_utils
 from cdp_backend.utils.file_utils import (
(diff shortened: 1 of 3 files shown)


evamaxfield (approved):
Looks good to me! Thanks! I am out right now so I will merge this now but can release later today in like 6 hours.

PR #237, opened 2023-06-04
Handle error in event index pipeline when a term is blank

conantp:
### Link to Relevant Issue

This pull request resolves an issue I identified and shared on Slack. 

### Description of Changes

This PR adds an if statement to generate_event_index_pipeline.py. When a sentence contains a "-" character, it winds up being evaluated as a "" value, which then causes an error on line 254:

```
target_term_index = terms.index(closest_term)
```

This causes the entire event index operation to fail, and the only option to restore index functionality is to remove the event that caused the issue. This PR allows event index to continue, skipping that term.

### DIFF
--- a/cdp_backend/pipeline/generate_event_index_pipeline.py
+++ b/cdp_backend/pipeline/generate_event_index_pipeline.py
@@ -248,6 +248,11 @@ def read_transcripts_and_generate_grams(
                             closest_term = term
                             closest_term_score = similarity
 
+                    # If the closest term is blank, skip processing this item
+                    # Otherwise this throws an error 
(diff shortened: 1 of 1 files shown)


evamaxfield (approved):
Looks good to me! Thanks!

I will push a new release in a bit. Unfortunately lost my two factor auth app for pypi so can't deploy new stuff right now hahaha. Already in the process for account recovery.

evamaxfield:
Oh woops, can you run `just lint` / fix the lint error please?

evamaxfield:
Will fix lint after merge! Thanks!

